PR #19, opened 2021-11-23
Update Token to be compatible with Governor DAO's

crazyrabbitLTC:
ERC20Votes is an extension from Openzeppelin that inherits both ERC20 and ERC20Permit, but adds the gas-efficient checkpointing functionality so that the token can be used natively chain by an OpenZeppelin Governor DAO. 

It should be a drop-in replacement for the previous inheritance of ERC20 and ERC20Permit but with more functionality. (I am doing this PR directly in the Github interface however so you might want to run the tests).

Learn more: 
https://docs.openzeppelin.com/contracts/4.x/api/governance

### DIFF
--- a/contracts/JBToken.sol
+++ b/contracts/JBToken.sol
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.6;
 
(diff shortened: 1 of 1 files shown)


mejango (approved):
sick thnx

PR #26, opened 2021-11-27
feat: add nounsDao ERC721

crazyrabbitLTC:
This PR replaces the stock OZ ERC721 with the NounsDAO checkpointable version of the ERC721. 

Why: 
Project ownership is denoted by ERC721's which are minted to a project owner's account. Assuming there is a future where the project owners constitute a "community" on juicebox, and are called upon to participate in governance actions, by replacing their ERC721 with the NounsDAO version we enable the project owners to collectively participate in governance via this contract which calculates the voting power of each address (counted as a number of projects they own). 

This PR only addresses making the ERC721 owner token that is minted compatible with the OZ Governor Ecosystem, meaning it will be ready for any Governor (or combination of governors) that is deployed in the future. This makes the JuiceBox protocol a bit more "future-ready".

### DIFF
--- a/contracts/JBProjects.sol
+++ b/contracts/JBProjects.sol
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.6;
 
-import '@openzeppelin/contracts/token/ERC721/ERC721.sol';
-
+import './abstract/ERC721Checkpointable.sol';
 import './abstract/JBOperatable.sol';
 import './interfaces/IJBProjects.sol';
 
@@ -15,7 +14,7 @@ import './libraries/JBOperations.sol';
   @dev
   Projects are represented as ERC-721's.
 */
-contract JBProjects is ERC721, IJBProjects, JBOperatable {
+contract JBProjects is ERC721Checkpointable, IJBProjects, JBOperatable {
   //*********************************************************************//
   // --------------------- private stored constants -------------------- //
   //*********************************************************************//
--- a/contracts/abstract/ERC721Checkpointable.sol
+++ b/contracts/abstract/ERC721Checkpointable.sol
@@ -1,0 +1,289 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/// @title Vote checkpointing for an ERC-721 token
+
+/*********************************
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ * ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ *
+ * ░░░░░░█████████░░█████████░░░ *
+ * ░░░░░░██░░░████░░██░░░████░░░ *
+ * ░░██████░░░████████░░░████░░░ *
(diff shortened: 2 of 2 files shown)


mejango:
I like the idea.

Problem is, there's no barrier to entry for Juicebox projects – anyone can launch one. 

In a future where project ownership has weight, we should consider adjusted metrics for representation. Any thoughts?

mejango:
Great perspective. Sounds good to me Dennison. 

@odd-amphora can you merge this once you've reviewed?

mejango:
O, actually one thing I just noticed.

ERC721 extends ERC721Enumerable. The mint function for Enumerable stores one extra reference to make the enumeration possible, which would make deploying projects slightly more expensive.

Let's tease out this tradeoff before merging.

cc: @crazyrabbitLTC

crazyrabbitLTC:
It's true, it's slightly more expensive, but it's a one-time cost and in
the scope of creating a juice box project, the time of day of your
deployment vis a vis gas costs would probably make a vastly larger
difference in the deployment costs.

But important to be cognizant about !

Dennison Bertram

www.dennisonbertram.com

NYC: +1 718 644 4272

mejango:
still torn on this one. it would in fact be tight if people could create governance strategies based on a collective of JB projects, 1 project 1 vote.

I'm sensitive to making deploying projects more expensive though.

crazyrabbitLTC:
What about making deploying projects as Clones of an established
architecture? That way it's a massive saving for deployment? (I actually
don't know if that's what is happening here, I'm on mobile so forgive me if
I'm totally off base)

What we are doing for our NFT DAO deployer tool is just cloning and
initializing the contracts. It's a massive gas savings:

https://github.com/daocity/create-nft-dao/blob/a11b26e9c697b15be9e5c949b0f5bb8bb10985ab/packages/hardhat/contracts/ERC721DAODeployer.sol#L84




Dennison Bertram

www.dennisonbertram.com

NYC: +1 718 644 4272

mejango:
@crazyrabbitLTC 
There isn't a separate contract that gets deployed each time someone launches a project, the tx just mints a new tokenId within the JBProjects NFT contract and does some funding cycle configuring. 
Projects can issue ERC-20 project tokens after the initial `launch` tx if they wish (expensive operation, I imagine could be improved using the clone / proxy strategy you're suggesting).

I believe this PR only effects the cost of `launch` by one storage op.

mejango:
@crazyrabbitLTC 

I think we should land this. We got rid of a bunch of SSTORE ops from the project launch operation by relying on ENS for handles instead of keeping our own handle registry (PR [here](https://github.com/jbx-protocol/juice-contracts-v2/pull/67)). Adding ERC721Enumerable makes sense if JBProjects is to be a standard that other ecosystems can use for project permissions.

cc @EzraWeller 

Should we add tests for the new abstract lib, or maybe a reference to the tests if they exist elsewhere?

mejango:
addressed in https://github.com/jbx-protocol/juice-contracts-v2/pull/67

PR #34, opened 2021-12-01
Optimization/reconfigure

mejango:
When a project is reconfiguring its funding cycles, most properties of the reconfiguration remain the same as the previous configuration.

Often times, only the `distributionLimit` is changing, but all metadata (duration, discount rates, currencies, redemption rates, overflow allowances) are staying the same.

This PR suggest a more efficient way to store some funding cycle properties that are assumed to change with less frequency.

The tradeoff is a slightly more expensive transaction to save rarely-updated metadata, and a slightly more convoluted pattern to read and grok.

### DIFF
--- a/contracts/JBController.sol
+++ b/contracts/JBController.sol
@@ -21,6 +21,8 @@ import './structs/JBFundingCycleData.sol';
 import './structs/JBFundingCycleMetadata.sol';
 import './structs/JBFundAccessConstraints.sol';
 import './structs/JBGroupedSplits.sol';
+import './structs/JBConfigurationCurrencyData.sol';
+import './structs/JBConfigurationOverflowAllowanceData.sol';
 
 // Inheritance
 import './interfaces/IJBController.sol';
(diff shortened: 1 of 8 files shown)


contracts/JBController.sol
```diff
@@ -202,6 +209,60 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
...
+  function currencyOf(
+    uint256 _projectId,
+    uint256 _configuration,
+    IJBTerminal _terminal
+  ) external view override returns (uint256) {
+    // Get the currency list for this project and this terminal.
+    JBConfigurationCurrencyData[] memory _data = _currencyDataOf[_projectId][_terminal];
+
+    // Loop through, starting with the latest.
+    for (uint256 _i = _data.length; _i > 0; _i--)
```
dhyon, line 231: Minor indexing stuff:
```suggestion
    for (uint256 _i = _data.length-1; _i >= 0; _i--)
```

contracts/JBController.sol
```diff
@@ -821,6 +882,8 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
         );
 
     // Set overflow allowances if there are any.
+    // The routine makes an assumption that `distributionLimitOf` changes with each reconfiguration
+    // much more frequently than `overflowAllowanceOf` or `currencyOf`. The data structures and storage patterns are thus different.
```
dhyon, line 886: Perhaps some analysis or polls to JB project owners can confirm these intuitions. That being said, I do believe that the distributionLimits are definitely prone to change from cycle to cycle.

contracts/JBController.sol
```diff
@@ -830,15 +893,43 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
...
+      ];
+
+      // Check if _constraints.currency exists and differs from latest value.
+      // If so, push the value into the array.
+      // This prevents needlessly storing currencies for each new reconfiguration if it isn't changing.
+      if (
+        (_constraints.currency > 0 && _currencyData.length == 0) ||
+        _currencyData[_currencyData.length].currency != _constraints.currency
+      )
+        _currencyDataOf[_projectId][_constraints.terminal].push(
```
dhyon, line 927: Very curious about the actual cost diff of pushing to an array infrequently vs. updating a map value everytime

contracts/structs/JBConfigurationFundingCycleMetadataData.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+struct JBConfigurationFundingCycleMetadataData {
+  // The configuration during which the new metadata was set.
+  uint56 configuration;
```
dhyon, line 6: Curious about the actual cost diff of storing as uint256 (since it's not packable anyways) and avoiding downsizing/typecasting costs vs. storing as a smaller uint56

dhyon (commented):
Interesting optimizations. Ultimately these changes need to be tested to gauge total gas consumption for the most common user flows, diffed against main.

mejango:
Archiving this.

PR #37, opened 2021-12-01
Feat: JBDirectory controller whitelist

drgorillamd:
image

Only non covered is actually covered :
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@ coverage/
 coverage.json
 
 *.txt
-.env
+
+#VSCode
 workspace.code-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,5 @@ coverage.json
 
 *.txt
 
-#VSCode
+# VSCode
 workspace.code-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ coverage/
 coverage.json
 
 *.txt
+.env
+workspace.code-workspace
--- a/contracts/JBDirectory.sol
+++ b/contracts/JBDirectory.sol
@@ -150,7 +150,7 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
     return IJBTerminal(address(0));
   }
 
-  function isKnownControllers(IJBController _address) public view override returns (bool) {
+  function isKnownController(IJBController _address) public view override returns (bool) {
     return _knownControllers[_address];
   }
 
@@ -191,9 +191,7 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
       projects.ownerOf(_projectId),
       _projectId,
       JBOperations.SET_CONTROLLER,
-      ( (address(controllerOf[_projectId]) == address(0) && msg.sender == address(_controller)) ||
-        address(controllerOf[_projectId]) == msg.sender )
-        && _knownControllers[_controller]
+      _knownControllers[_controller]
     )
   {
     // Can't set the zero address.
@@ -345,11 +343,8 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
     @param _controller The terminal to be added.
   */
   function addKnownController(IJBController _controller) external override onlyOwner {
-    // Can't add zero address.
-    require(_controller != IJBController(address(0)), '0x30: ZERO_ADDRESS');
-
     // Check that the controller has not already been added.
-    require(!_knownControllers[_controller], '0x31: ALREADY_ADDED');
+    require(!_knownControllers[_controller], '0x30: ALREADY_ADDED');
 
     // Add the controller to the list of known controllers.
     _knownControllers[_controller] = true;
@@ -365,7 +360,7 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
   */
   function removeKnownController(IJBController _controller) external override onlyOwner {
     // Not in the known controllers list
-    require(_knownControllers[_controller], '0x33: NOT_FOUND');
+    require(_knownControllers[_controller], '0x31: NOT_FOUND');
 
     // Remove the controller from the list of known controllers.
     delete _knownControllers[_controller];
--- a/contracts/JBDirectory.sol
+++ b/contracts/JBDirectory.sol
@@ -10,9 +10,7 @@ import './libraries/JBOperations.sol';
 
 /**
   @notice
-  Keeps a reference of which terminal contracts each project is currently
-  accepting funds through, and which controller contract is managing each
-  project's tokens and funding cycles.
+  Keeps a reference of which terminal contracts each project is currently accepting funds through, and which controller contract is managing each project's tokens and funding cycles.
 */
 contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
   //*********************************************************************//
@@ -38,9 +36,7 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
 
   /**
     @notice
-    Addresses that can change a project's controllers. Those are controllers
-    that have been vetted and verified by Juicebox owners as well as custom
-    launching project contracts
+    Addresses that can set a project's controller. These addresses/contracts have been vetted and verified by Juicebox owners.
    */
   mapping(address => bool) private _setControllerAllowlist;
 
@@ -154,7 +150,7 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
     return IJBTerminal(address(0));
   }
 
-  /*
+  /**
     @notice
     Whether or not a specified address is allowed to set controllers.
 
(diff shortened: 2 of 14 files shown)


drgorillamd:
REVIEW/DISCUSS access for `setControllerOf(...)` :
Now, any address can call `setControllerOf` and, as long as `IJBController _controller` is in knownControllers, `requirePermissionAllowingOverride` will be overridden (without, for instance, the funding cycle check and the premigration process our canonical controller is doing)

Should :
```
    requirePermissionAllowingOverride(
      projects.ownerOf(_projectId),
      _projectId,
      JBOperations.SET_CONTROLLER,
      _knownControllers[_controller]
    )
```
    
Rather be : 
```
    requirePermissionAllowingOverride(
      projects.ownerOf(_projectId),
      _projectId,
      JBOperations.SET_CONTROLLER,
      _knownControllers[**msg.sender**]
    )
```
(the knownController calling it being the "new" or current one of a project, so the premigration&migration are done as intended)

contracts/JBDirectory.sol
```diff
@@ -182,27 +182,21 @@ contract JBDirectory is IJBDirectory, JBOperatable, Ownable {
     @param _projectId The ID of the project to set a new controller for.
     @param _controller The new controller to set.
   */
-  function setControllerOf(uint256 _projectId, IJBController _controller)
-    external
-    override
-    requirePermissionAllowingOverride(
```
drgorillamd, line 178: extra emphasis on this change:)

.gitignore
```diff
@@ -13,4 +13,6 @@ deployments/localhost/**
 coverage/
 coverage.json
 
-*.txt
\ No newline at end of file
+*.txt
+.env
```
odd-amphora, line 17: .env already exists at the top.

Maybe separate this by

#VSCode
workspace.code-workspace

contracts/JBDirectory.sol
```diff
@@ -1,16 +1,20 @@
...
 import './interfaces/IJBDirectory.sol';
 import './abstract/JBOperatable.sol';
 import './libraries/JBOperations.sol';
 
 /**
   @notice
-  Keeps a reference of which terminal contracts each project is currently accepting funds through, and which controller contract is managing each project's tokens and funding cycles.
+  Keeps a reference of which terminal contracts each project is currently
+  accepting funds through, and which controller contract is managing each
+  project's tokens and funding cycles.
```
odd-amphora, line 15: Why this change? Do we have a set column width? (we should)

contracts/JBDirectory.sol
```diff
@@ -32,6 +36,14 @@ contract JBDirectory is IJBDirectory, JBOperatable {
   */
   mapping(uint256 => mapping(address => IJBTerminal)) private _primaryTerminalOf;
 
+  /**
```
odd-amphora, line 39: Per my comment above, the line wrap/column width seems inconsistent.

contracts/JBDirectory.sol
```diff
@@ -32,6 +36,14 @@ contract JBDirectory is IJBDirectory, JBOperatable {
...
+    @notice
+    Addresses that can change a project's controllers. Those are controllers
+    that have been vetted and verified by Juicebox owners as well as custom
+    launching project contracts
```
odd-amphora, line 43: Consider: Addresses that can set a project's controller. These addresses/contracts have been vetted and verified by Juicebox owners.

contracts/JBDirectory.sol
```diff
@@ -154,43 +178,31 @@ contract JBDirectory is IJBDirectory, JBOperatable {
...
     Update the controller that manages how terminals interact with the ecosystem.
-
     @dev 
     A controller can be set if:
-    - the message sender is the project owner or an operator is changing the controller.
-    - or, the controller hasn't been set yet and the message sender is the controller being set.
-    - or, the current controller is setting a new controller.
-
+    - the message sender is the project owner or an operator having the correct authorization.
+    - or an allowed address is setting a new controller.
```
odd-amphora, line 187: nit) or,

contracts/JBDirectory.sol
```diff
@@ -319,4 +331,42 @@ contract JBDirectory is IJBDirectory, JBOperatable {
...
+
+  /** 
+    @notice
+    The owner (Juicebox multisig) can add addresses which are allowed to change
+    a project's controller. Those addresses are known and vetted controllers as well as
+    contracts designed to launch new projects. This is not a requirement for all controllers.
+    However, unknown controllers may require additional transactions to perform certain operations.
+
+    @dev
+    If you would like your JBController allowlisted, please reach out to the Juicebox dev team.
```
odd-amphora, line 343: If you would like an address/contract allowlisted

^ remember these are no longer necessarily controllers

test/jb_directory/add_known_controller.test.js
```diff
@@ -0,0 +1,64 @@
...
+      jbDirectory.connect(deployer).addToSetControllerAllowlist(mockJbController.address)
+    ).to.emit(jbDirectory, 'AddToSetControllerAllowlist')
+      .withArgs(mockJbController.address, deployer.address);
+
+    expect(
+      await jbDirectory.isAllowedToSetController(mockJbController.address)
+    ).to.be.true;
+  });
+
+  it('Can\'t add known controller if caller is not JBDirectory owner', async function () {
```
odd-amphora, line 46: If \ characters are in the name, you will need to wrap with ` ` quotes - otherwise the test name will show as, "Can"

contracts/JBDirectory.sol
```diff
@@ -154,43 +178,31 @@ contract JBDirectory is IJBDirectory, JBOperatable {
...
   function setControllerOf(uint256 _projectId, IJBController _controller)
     external
     override
     requirePermissionAllowingOverride(
       projects.ownerOf(_projectId),
       _projectId,
       JBOperations.SET_CONTROLLER,
-      (address(controllerOf[_projectId]) == address(0) && msg.sender == address(_controller)) ||
-        address(controllerOf[_projectId]) == msg.sender
+      (_setControllerAllowlist[address(_controller)] && _setControllerAllowlist[msg.sender])
```
odd-amphora, line 198: Should we use `isAllowedToSetController` in these cases?

test/jb_directory/add_known_controller.test.js
odd-amphora: Fix in all files in this PR.

test/jb_directory/add_known_controller.test.js
```diff
@@ -0,0 +1,64 @@
...
+    const { deployer, jbDirectory, mockJbController } = await setup();
+
+    await jbDirectory.connect(deployer).addToSetControllerAllowlist(mockJbController.address)
+
+    await expect(
+      jbDirectory.connect(deployer).addToSetControllerAllowlist(mockJbController.address)
+    ).to.revertedWith('0x30: ALREADY_ADDED');
+  });
+
+});
```
odd-amphora, line 64: Let's add a test that adds and then removes controller(s) and ensures they are no longer allowed.

odd-amphora: eh, i forgot about the tests below, nvm.

contracts/JBDirectory.sol
drgorillamd: isAllowedToSetController is a public function -> would slightly increase gas consumption (alt being a second internal function overloading it, costing then 1U for the jump + deployment overhead, but I guess that's not really necessary here)

test/jb_directory/add_known_controller.test.js
drgorillamd: the \ is to escape the ' character, display is ok but double-checking other tests, this is indeed not consistent across them

image

.gitignore
```diff
@@ -13,4 +13,7 @@ deployments/localhost/**
 coverage/
 coverage.json
 
-*.txt
\ No newline at end of file
+*.txt
+
+#VSCode
```
odd-amphora, line 18: super nit) space in between `#`

# VSCode

test/jb_directory/add_set_controller_allowed_list.test.js
```diff
@@ -0,0 +1,64 @@
+import { expect } from 'chai';
+import { ethers } from 'hardhat';
+
+import { deployMockContract } from '@ethereum-waffle/mock-contract';
+
+import jbController from '../../artifacts/contracts/interfaces/IJBController.sol/IJBController.json';
+import jbOperatoreStore from '../../artifacts/contracts/JBOperatorStore.sol/JBOperatorStore.json';
+import jbProjects from '../../artifacts/contracts/JBProjects.sol/JBProjects.json';
+
+describe('JBDirectory::AddToSetControllerAllowlist(...)', function () {
```
odd-amphora, line 10: super nit) camel case `addToSetControllerAllowlist` (replace in other files too)

odd-amphora (approved):

mejango (approved):
Looks good, i like this pattern.

Will projects be able to add custom controllers for themselves?

PR #41, opened 2021-12-13
JBProject tests

odd-amphora:
Screen Shot 2021-12-14 at 11 38 19 AM

### DIFF
--- a/contracts/abstract/JBProject.sol
+++ b/contracts/abstract/JBProject.sol
@@ -5,6 +5,7 @@ import '@openzeppelin/contracts/access/Ownable.sol';
 import '@openzeppelin/contracts/utils/Address.sol';
 
 import './../interfaces/IJBDirectory.sol';
+import './../libraries/JBTokens.sol';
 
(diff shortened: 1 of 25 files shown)


contracts/abstract/JBProject.sol
```diff
@@ -29,7 +29,7 @@ abstract contract JBProject is Ownable {
   }
 
   /** 
-      Received funds go streight to the project.
+      Received funds go straight to the project.
     */
   receive() external payable {
     _pay(msg.sender, '', false, address(0));
```
mejango, line 35: i think the token representation of ETH is no longer 0x00. 

https://github.com/jbx-protocol/juice-contracts-v2/blob/81c78393a4ff46dfcd354beeaa771f2516b2a7ad/contracts/JBETHPaymentTerminal.sol#L92

mejango (commented):
word. nice. small bug then lfgtm

mejango (approved):
cool this looks very gtm

PR #43, opened 2021-12-15
JBETHPaymentTerminalStore tests and fixes

dhyon:
I made some small changes to the contracts as well.

Screen Shot 2021-12-31 at 6 20 47 PM


contracts/JBETHPaymentTerminal.sol
```diff
@@ -723,7 +723,7 @@ contract JBETHPaymentTerminal is
       msg.sender,
       _amount,
       _projectId,
-      (_preferClaimedTokens ? 1 : 0) | uint160(_beneficiary),
+      (_preferClaimedTokens ? 1 : 0) | (uint160(_beneficiary) << 1),
```
dhyon, line 726: @mejango

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -282,7 +282,7 @@ contract JBETHPaymentTerminalStore {
         _weightedAmount,
         address(uint160(_preferClaimedTokensAndBeneficiary >> 1)),
         'ETH received',
-        (_preferClaimedTokensAndBeneficiary & 1) == 0,
+        (_preferClaimedTokensAndBeneficiary & 1) == 1,
```
dhyon, line 285: @mejango

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -201,7 +201,7 @@ contract JBETHPaymentTerminalStore {
     @dev
     Mint's the project's tokens according to values provided by a configured data source. If no data source is configured, mints tokens proportional to the amount of the contribution.
 
-    @dev
```
odd-amphora, line 204: It may be worth making this formatting change separately against main, getting it merged, and then rebasing your changes on top of it to make this diff easier.

contracts/mocks/JBFundingCycleDataSourceMock.sol
```diff
@@ -0,0 +1,30 @@
...
+    view
+    override
+    returns (
+      uint256 amount,
+      string memory memo,
+      IJBRedemptionDelegate delegate,
+      bytes memory delegateMetadata
+    )
+  {}
+}
```
odd-amphora, line 30: As mentioned over Discord, we should need to create mock contracts for interfaces. You can call `deployMockContract` against the interface ABI. Let's remove these. JBProject was different since it's not an interface and is never extended by any of the core JB contracts. The intent is for projects to extend this themselves.

contracts/mocks/JBPayDelegateMock.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import './../interfaces/IJBPayDelegate.sol';
+
+contract JBPayDelegateMock is IJBPayDelegate {
+  function didPay(JBDidPayData calldata _param) external override {}
+}
```
odd-amphora, line 8: Same here.

test/jb_eth_payment_terminal_store/record_payment_from.test.js
```diff
@@ -0,0 +1,537 @@
...
+import { packFundingCycleMetadata } from '../helpers/utils';
+
+import jbController from '../../artifacts/contracts/JBController.sol/JBController.json';
+import jbDirectory from '../../artifacts/contracts/JBDirectory.sol/JBDirectory.json';
+import jBFundingCycleStore from '../../artifacts/contracts/JBFundingCycleStore.sol/JBFundingCycleStore.json';
+import jbFundingCycleDataSource from '../../artifacts/contracts/mocks/JBFundingCycleDataSourceMock.sol/JBFundingCycleDataSourceMock.json';
+import jbPayDelegate from '../../artifacts/contracts/mocks/JBPayDelegateMock.sol/JBPayDelegateMock.json';
+import jbPrices from '../../artifacts/contracts/JBPrices.sol/JBPrices.json';
+import jbProjects from '../../artifacts/contracts/JBProjects.sol/JBProjects.json';
+import jbTokenStore from '../../artifacts/contracts/JBTokenStore.sol/JBTokenStore.json';
```
odd-amphora, line 15: We should be able to use ../../artifacts/contracts/interfaces/... for all of these.

test/jb_eth_payment_terminal_store/record_payment_from.test.js
```diff
@@ -0,0 +1,537 @@
...
+      mockJbFundingCycleStore,
+      jbEthPaymentTerminalStore,
+      timestamp,
+    } = await setup();
+
+    // Set terminal address
+    await jbEthPaymentTerminalStore.claimFor(terminal.address);
+
+    const reservedRate = 0;
+    const packedFCM = packFundingCycleMetadata({ pausePay: 0, reservedRate: reservedRate });
```
odd-amphora, line 85: nit) consider: metadata here and elsewhere. I'm not a fan of acronyms personally.

odd-amphora (changes requested):

contracts/JBETHPaymentTerminalStore.sol
dhyon: I like this idea

https://github.com/jbx-protocol/juice-contracts-v2/pull/44

contracts/JBETHPaymentTerminal.sol
dhyon: @drgorillamd

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -403,8 +400,7 @@ contract JBETHPaymentTerminalStore {
     }
 
     // Convert the amount to wei.
-    // A currency of 0 should be interpreted as whatever the currency being distributed is.
```
dhyon, line 406: This seemed a bit tribal to me so I generalized the logic to skip if the preferred currency is ETH; otherwise we do the price conversion.

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -462,25 +458,27 @@ contract JBETHPaymentTerminalStore {
       revert CURRENCY_MISMATCH();
     }
 
-    // Convert the amount to wei.
-    // A currency of 0 should be interpreted as whatever the currency being withdrawn is.
-    withdrawnAmount = _currency == 0
-      ? _amount
-      : PRBMathUD60x18.div(_amount, prices.priceFor(_currency, JBCurrencies.ETH));
+    uint256 _newOverflowAllowanceOf = usedOverflowAllowanceOf[_projectId][
```
dhyon, line 461: Decided to calculate this once here since this value is used effectively twice later on

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -772,13 +777,13 @@ contract JBETHPaymentTerminalStore {
...
-      ? 0 // Get the current price of ETH. // A currency of 0 should be interpreted as whatever the currency being withdrawn is.
-      : _currency == 0
+      ? 0
+      : (_currency == JBCurrencies.ETH)
       ? _distributionRemaining
       : PRBMathUD60x18.div(_distributionRemaining, prices.priceFor(_currency, JBCurrencies.ETH));
 
     // Overflow is the balance of this project minus the amount that can still be distributed.
-    return _balanceOf < _ethDistributionRemaining ? 0 : _balanceOf - _ethDistributionRemaining;
+    return _balanceOf <= _ethDistributionRemaining ? 0 : _balanceOf - _ethDistributionRemaining;
```
dhyon, line 786: conceptually the same but we save ourselves from doing subtraction in the case balanceOf == _ethDistributionRemaining

contracts/JBETHPaymentTerminalStore.sol
mejango: nice

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -419,9 +415,9 @@ contract JBETHPaymentTerminalStore {
     }
 
     // Store the new amount.
-    usedDistributionLimitOf[_projectId][fundingCycle.number] = _newUsedDistributionLimitOf;
+    usedDistributionLimitOf[_projectId][fundingCycle.configuration] = _newUsedDistributionLimitOf;
```
mejango, line 418: I think here we want .number

mejango: the distributionLimit is a per-funding cycle basis, not a per-configuration basis. Projects should get the limit again as funding cycles roll over without explicit reconfiguration.

overflowAllowance, on the other hand, i think might be better as per configuration. i.e. "the DAO can access 30 ETH from overflow" and it doesnt reset each funding cycle.

contracts/JBETHPaymentTerminalStore.sol
mejango: maybe `_newUsedOverflowAllowanceOf`

mejango: nice change

contracts/JBETHPaymentTerminalStore.sol
mejango: nice

mejango (commented):
love it. one note regarding .number vs .configuration thats worth settling.

contracts/JBETHPaymentTerminalStore.sol
dhyon: Oh yeah now that I think back this was a cause of confusion for me. Definitely makes sense to look up distributionLimit on a per-funding cycle basis.

dhyon: Made the change from funding cycle configuration to number here but the controller and terminal that @drgorillamd is handling will need some updates. 

controller.remainingDistributionLimitOf method
terminal.distributionLimitOf mapping
terminal._configure method

will point those out in the respective PRs

odd-amphora (approved):

PR #45, opened 2021-12-15
Remove one time funding cycles 

mejango:
These haven't yet been used productively, and have only caused accidental misconfigurations.
I can't imagine a particularly productive use case for this given the campaigns we've empirically seen over the past 5 months of JB projects in the wild.

I suggest we remove it, which will also remove unneeded complexity.

### DIFF
--- a/contracts/JBFundingCycleStore.sol
+++ b/contracts/JBFundingCycleStore.sol
@@ -12,6 +12,16 @@ import './abstract/JBControllerUtility.sol';
 */
 contract JBFundingCycleStore is JBControllerUtility, IJBFundingCycleStore {
   //*********************************************************************//
+  // ------------------------- private constants ----------------------- //
+  //*********************************************************************//
+
+  /** 
+    @notice
+    A funding cycle's discount rate is expressed as a percentage out of 100000000.
+  */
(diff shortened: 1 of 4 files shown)


contracts/JBFundingCycleStore.sol
```diff
@@ -256,8 +249,8 @@ contract JBFundingCycleStore is JBControllerUtility, IJBFundingCycleStore {
     // Duration must fit in a uint64, and must be greater than 1000 seconds to prevent manipulative miner behavior.
     require(_data.duration <= type(uint64).max && _data.duration > 1000, '0x15: BAD_DURATION');
 
-    // Discount rate token must be less than or equal to 100%. A value of 1000000001 means non-recurring.
-    require(_data.discountRate <= 1000000001, '0x16: BAD_DISCOUNT_RATE');
+    // Discount rate token must be less than or equal to 100%.
+    require(_data.discountRate <= 100000000, '0x16: BAD_DISCOUNT_RATE');
```
odd-amphora, line 253: Consider extracting this into a meaningful constant, MAX_DISCOUNT_RATE, for better readability. I see it's used in many places throughout file, so we should replace them there with the new constant.

mejango: done

PR #47, opened 2021-12-24
Initial Foundry setup

odd-amphora:


dhyon (approved):
this seems stable enough to work off of

PR #52, opened 2022-01-05
JBFundingCycleMetadataResolver lib tests

dhyon:
Screen Shot 2022-01-05 at 6 40 53 PM

### DIFF
--- a/contracts/libraries/JBFundingCycleMetadataResolver.sol
+++ b/contracts/libraries/JBFundingCycleMetadataResolver.sol
@@ -144,6 +144,6 @@ library JBFundingCycleMetadataResolver {
     // use redeem data source in bit 67.
     if (_metadata.useDataSourceForRedeem) packed |= 1 << 67;
     // data source address in bits 68-227.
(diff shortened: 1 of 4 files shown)


contracts/libraries/JBFundingCycleMetadataResolver.sol
```diff
@@ -144,6 +144,6 @@ library JBFundingCycleMetadataResolver {
     // use redeem data source in bit 67.
     if (_metadata.useDataSourceForRedeem) packed |= 1 << 67;
     // data source address in bits 68-227.
-    packed |= uint160(address(_metadata.dataSource)) << 68;
+    packed |= uint256(uint160(address(_metadata.dataSource))) << 68;
```
dhyon, line 147: cast into uint256 before shifting or we lose the bits

mejango: nice catch

odd-amphora (approved):

PR #53, opened 2022-01-07
Remove superfluous testing

odd-amphora:
Code cov already executes tests, compiles, and brings up a local chain – so we're effectively running all tests twice. We can remove all this noise.


mejango (approved):
lg

PR #59, opened 2022-01-11
feature/fee-gauge

mejango:
Better channel payment payer to the data source to be able to make decisions based on who is paying (including if its a particular payment terminal which is sending a fee)

image


contracts/JBETHPaymentTerminal.sol
```diff
@@ -694,8 +721,19 @@ contract JBETHPaymentTerminal is
     address _beneficiary,
     string memory _memo
   ) private returns (uint256 feeAmount) {
+    // Get the fee discount.
+    uint256 _feeDiscount = feeGuage == IJBFeeGuage(address(0))
+      ? 0
+      : feeGuage.currentDiscountFor(_projectId);
+
+    // Set the discounted fee if its valid.
+    if (_feeDiscount > 1000000) _feeDiscount = 0;
```
mejango, line 730: need to define the const for MAX_FEE_DISCOUNT

drgorillamd (approved):
Small changes for a whole new universe of possibilities B-)

contracts/JBETHPaymentTerminal.sol
drgorillamd: @mejango why not` > 1000000) _feeDiscount = 1000000;` ?

mejango: yeah that works also. 

Make sure to pull this out into a const in the lib too plz!!

mejango: also @drgorillamd might as well pump up the fidelity of the const. I think i just plopped a random number when i was doing it.

drgorillamd: done, same granularity (10**9) - I kepr the '=0', if there is a gauge bug, maybe easier to give back extra fee perceived than to ask for such ':)

PR #60, opened 2022-01-12
Minimal changes for a composable projects registry

EzraWeller:
Hey JB crew! 

This PR comes from a discussion and help I got from Jango in Discord the past few days.

### Goal
The need for an on-chain registry of projects (in the vague sense) is wider than just Juicebox. At [R Group](https://rrrrr.group/), we're planning to build a few upcoming projects that need one, and I know projects like clr.fund and gitcoin grants could also use one. 

I like JB v2's approach, representing projects as NFTs, and so I wanted to see if we could make JB v2's projects contract into a project registry that could easily be reused outside of JB, without disrupting what JB is already doing.

### Changes
- JBController, `launchProjectFor` -- adds the possibility for this to be called for a handle that already exists by the owner of that handle's project, making it convenient to create a Juicebox DAO for an already existing project. This makes sense if the project registry is a more independent thing, IMO, and it's a small change.
(I didn't see tests for the controller in the repo, so didn't add tests for this yet -- did I just overlook them?)

- JBProjects -- made ownable and separated out the calculation of challenge expiries into a new contract (JBExpirySource) that the owner of JBProjects can replace. If the projects registry does start getting used across projects, I'd expect more "hoarding" of handles to start happening. This provides flexible recourse without introducing major changes or a super wide governance surface. Not 100% sure of this approach, but it is interesting.

### DIFF
--- a/contracts/JBController.sol
+++ b/contracts/JBController.sol
@@ -35,6 +35,7 @@ import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
 //*********************************************************************//
 error BURN_PAUSED_AND_SENDER_NOT_VALID_TERMINAL_DELEGATE();
 error CALLER_NOT_CURRENT_CONTROLLER();
+error CALLER_NOT_PROJECT_OWNER();
 error CANT_MIGRATE_TO_CURRENT_CONTROLLER();
 error CHANGE_TOKEN_NOT_ALLOWED();
 error INVALID_BALLOT_REDEMPTION_RATE();
@@ -321,8 +322,19 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
       revert INVALID_BALLOT_REDEMPTION_RATE();
     }
 
-    // Create the project for into the wallet of the message sender.
-    projectId = projects.createFor(_owner, _handle, _metadataCid);
+    // Get the projectId (will be 0 if project doesn't exist)
+    projectId = projects.idFor(_handle);
+
+    // If project already exists and msg.sender is not the project owner, revert
+    // (if project already exists and msg.sender is the project owner, continue without creating a new project).
+    if (projectId != 0 && msg.sender != projects.ownerOf(projectId)) {
+      revert CALLER_NOT_PROJECT_OWNER();
+    }
+
+    // If project doesn't already exist, create it
+    if (projectId == 0) {
+      projectId = projects.createFor(_owner, _handle, _metadataCid);
+    }
 
     // Set the this contract as the project's controller in the directory.
     directory.setControllerOf(projectId, this);
--- a/contracts/JBExpirySource.sol
+++ b/contracts/JBExpirySource.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.6;
 
-import '../interfaces/IJBExpirySource.sol';
+import './interfaces/IJBExpirySource.sol';
 
-abstract contract JBExpirySource is IJBExpirySource {
+contract JBExpirySource is IJBExpirySource {
   /** 
     @notice
     The number of seconds in 365 days.
@@ -17,7 +17,7 @@ abstract contract JBExpirySource is IJBExpirySource {
     (e.g. to help recover specific handles)
   */
 
-  function getExpiryFor(uint256) external view override returns (uint256 challengeExpiry) {
+  function getExpiryFor(uint256) external view override returns (uint256) {
     return block.timestamp + _SECONDS_IN_YEAR;
   }
 }
--- a/contracts/JB1YearChallengePeriodSource.sol
+++ b/contracts/JB1YearChallengePeriodSource.sol
@@ -5,12 +5,6 @@ import './interfaces/IJBChallengePeriodSource.sol';
 
 contract JB1YearChallengePeriodSource is IJBChallengePeriodSource {
   /** 
-    @notice
-    The number of seconds in 365 days.
-  */
-  uint256 private constant _SECONDS_IN_YEAR = 31536000;
-
-  /** 
     @notice Get the total challenge expiry time when challenging an inactive project handle.
 
     @dev The projectId param is unused here, but could be used in other implementations of IJBChallengeExpiry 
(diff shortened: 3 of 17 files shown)


contracts/JBController.sol
```diff
@@ -321,8 +322,19 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
...
+    // If project already exists and msg.sender is not the project owner, revert
+    // (if project already exists and msg.sender is the project owner, continue without creating a new project).
+    if (projectId != 0 && msg.sender != projects.ownerOf(projectId)) {
+      revert CALLER_NOT_PROJECT_OWNER();
+    }
+
+    // If project doesn't already exist, create it
+    if (projectId == 0) {
+      projectId = projects.createFor(_owner, _handle, _metadataCid);
+    }
```
mejango, line 337: I like this change.

contracts/JBExpirySource.sol
```diff
@@ -0,0 +1,23 @@
...
+  uint256 private constant _SECONDS_IN_YEAR = 31536000;
+
+  /** 
+    @notice Get the total challenge expiry time when challenging an inactive project handle.
+
+    @dev The projectId param is unused here, but could be used in other implementations of IJBChallengeExpiry 
+    (e.g. to help recover specific handles)
+  */
+
+  function getExpiryFor(uint256) external view override returns (uint256) {
```
mejango, line 20: `getChallengePeriod"?

maybe the `expiry` metaphor is clicking for you well, im happy to go with that instead of the `challengePeriod` metaphor, but just wanna make sure.

`challengePeriod` to me suggests the time the the owner of the handle has to answer the challenge by renewing.

contracts/JBExpirySource.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import './interfaces/IJBExpirySource.sol';
+
+contract JBExpirySource is IJBExpirySource {
```
mejango, line 6: JB1YearChallengePeriodSource  and IJBChallengePeriodSource?

mejango (commented):
real interesting changes. I like this extensibility.

wondering if we should hardcode a minimum `challengePeriod` of say 14 days. Prevents the `owner` from providing a hyper annoying source of 1 second or whatever.

@drgorillamd the changes to JBController might nudge some tests in progress that haven't been merged to main yet.

drgorillamd:
Nice!
Tests for JBController are done and in a PR right now (waiting for other approval and correcting some minor issues/typo/etc), happy to write what's needed (quick overview, that's like 2 extra tests, no biggie:)

contracts/JBExpirySource.sol
EzraWeller: Down for either, but my inclination is either `expiry` everywhere or `period` everywhere. Should we do period?

mejango: cool. `period` or `challengePeriod`?

leaning towards `challengePeriod` cus `expiry` and `JBExpiry` is a little non-descript.

EzraWeller:
> real interesting changes. I like this extensibility.
> 
> wondering if we should hardcode a minimum `challengePeriod` of say 14 days. Prevents the `owner` from providing a hyper annoying source of 1 second or whatever.
> 
> @drgorillamd the changes to JBController might nudge some tests in progress that haven't been merged to main yet.

Actually, it was kinda my intention that the owner be able to set a really short expiry, because that's their way of forcing a bad actor into giving up a handle (make the expiry time short enough that's it's not worth holding on to, then have it revert to a regular expiry transfer). Since we're giving the owner a lot of trust here, I think it's okayto just trust them and not have a minimum, but I'm not opposed to a short minimum either. 14 days seems kinda long for that to me, maybe just 1 day? WDYT?

contracts/JBExpirySource.sol
EzraWeller: sounds good, will update it

EzraWeller:
Small note: I haven't made a unit test for `JB1YearChallengePeriodSource.sol` since it's so simple, but can make one if needed.

mejango:
> > real interesting changes. I like this extensibility.
> > wondering if we should hardcode a minimum `challengePeriod` of say 14 days. Prevents the `owner` from providing a hyper annoying source of 1 second or whatever.
> > @drgorillamd the changes to JBController might nudge some tests in progress that haven't been merged to main yet.
> 
> Actually, it was kinda my intention that the owner be able to set a really short expiry, because that's their way of forcing a bad actor into giving up a handle (make the expiry time short enough that's it's not worth holding on to, then have it revert to a regular expiry transfer). Since we're giving the owner a lot of trust here, I think it's okayto just trust them and not have a minimum, but I'm not opposed to a short minimum either. 14 days seems kinda long for that to me, maybe just 1 day? WDYT?

Got it, that makes sense to. I suppose this is the last thing remaining imo: does this solution cleanly cover all the cases we want to cover to solve this problem effectively?

The flow would be: 
- anyone can challenge a handle at any time, which will start a challenge period of arbitrary length.
- the current owner of the handle must send a tx to renew the tx during the challenge period, else someone can claim the handle away from the current holder.

EzraWeller:
I'd add / adjust:
> * anyone can challenge an _unchallenged_ handle at any time, which will start a challenge period of arbitrary length.
> * anyone can claim a handle if it is not renewed during a challenge period
> * owner of the projects contract sets the logic for how challenge periods are calculated at the time of the challenge, and can swap this out anytime (won't affect challenges in progress).

Some consequences of this flow are:
- handles are free to claim, which means there's probably untapped economic potential being left on the table here (see ENS's pricing / renting scheme, etc.)
- since there's no way to ensure a specific person claims a handle after a successful challenge, this mechanic mostly solves the problem of handles getting stuck in the hands of inactive projects
- however, if needed it also gives the contract owner some ability to force addresses to give up specific handles by giving them very short challenge periods

Some questions could be:
- is it okay that specific transfers can't be enforced? would it ever be important to guarantee a specific project ends up with a handle? E.g. Uniswap DAO makes a JB treasury and they want the Uniswap handle, but they can't ever get it?
- is it okay that anyone can claim initial handles for free? JB v1 projects won't be guaranteed the same handle in v2, right? Is that okay?

EzraWeller:
Closing in favor of the new approach

PR #63, opened 2022-01-19
feat: forge

tarrencev:
adds foundry to support solidity/fuzz testing. right now, this breaks `hardhat compile`, if we're interested in merging it, we can try to find a solution to that

### DIFF
--- a/contracts/test/fuzz/JBOperatorStore.t.sol
+++ b/contracts/test/fuzz/JBOperatorStore.t.sol
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.6;
+
+import 'ds-test/test.sol';
+import 'forge-std/Vm.sol';
+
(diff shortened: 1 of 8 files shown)


contracts/test/fuzz/JBOperatorStore.t.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.6;
+
+import 'ds-test/test.sol';
+import 'forge-std/Vm.sol';
+
+import {JBOperatorStore, PERMISSION_INDEX_OUT_OF_BOUNDS} from '../../../contracts/JBOperatorStore.sol';
+import {JBOperatorData} from '../../../contracts/structs/JBOperatorData.sol';
+
+contract JBOperatorStoreTest is DSTest {
```
tarrencev, line 10: Don't have a good enough handle on the code base to write meaningful tests right now, but included this as an example incase it is useful (although prob not testing anything useful)

mejango:
can this be scrapped @drgorillamd ?

PR #65, opened 2022-01-19
project metadata mapped to domains

mejango:
Projects can now set a metadataCid for any number of domain namespaces

### DIFF
--- a/contracts/structs/JBProjectMetadataData.sol
+++ b/contracts/structs/JBProjectMetadataData.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+struct JBProjectMetadataData {
+  // An IPFS content ID where the metadata is found.
+  string cid;
+  // The domain within which the metadata applies.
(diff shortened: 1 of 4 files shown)


contracts/structs/JBProjectMetadataData.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+struct JBProjectMetadataData {
```
odd-amphora, line 4: nit) why not just JBProjectMetadata? Similar to JBFundingCycleMetadata

odd-amphora (approved):

odd-amphora (changes requested):
Can we get the tests fixed before merge?

contracts/structs/JBProjectMetadataData.sol
mejango: yay

dhyon (approved):
LGTM

odd-amphora (approved):

PR #67, opened 2022-01-21
Remove handles so we can use ENS (more reusable project NFTs)

EzraWeller:
The idea here is to make the project NFT system for JB v2 more re-usable by other projects, so that it might act like an interoperable project registry in the future (we're interested in building on it).

JBController -- adds 'launchFundingCycleFor`, a way to launch funding cycles for already existing projects. Ideally the JB frontend can combine this flow with the project launch flow, so you can either launch a project + funding cycle together (as now), or select an existing project you own to launch for.

JBProjects + others -- removed `handles` completely so we can use ENS for project names instead. These contracts don't need to know anything about ENS for this to work, as the links between projects and names can be done with ENS' reverse and text records. I think ENS is the way to go here because 1) JB contracts can simplify, 2) ENS already deals with the difficult problem of distributing names somewhat fairly and safely.

Here's roughly how we're suggesting project names to be handled with ENS:
```
---Project to ENS name---
projectId —> project owner —> does the project owner have a reverse record ENS name?
- Yes --> do the ENS name’s text records specify an ENS name for this projectId?
  - Yes —> does the project owner own that ENS name? 
    - Yes —> you found it
    - No —> invalid, use projectId as fallback name
  - No —> treat the reverse record ENS name as the project name
- No --> invalid, use projectId as fallback name
```
```
---ENS name to project---
Does the ENS name contain a projectId in the “Juicebox” text record?
- Yes —> does the project owner own this ENS name?
  - Yes —> is the ENS name the project owner’s reverse record or does their reverse record confirm this ENS name is signed to that projectId?
    - Yes —> you found it
    - No —> invalid, use projectId as fallback name
  - No —> invalid, use projectId as fallback name
- No —> no associated Juicebox project
```

To make this work, we'll need to give everyone instructions for setting up their name in ENS, as well as deal with name look ups / reverse look ups in the front end. It could make sense to work on some libraries for that next to encourage others to re-use the registry.

Hope that all makes sense!


contracts/JBController.sol
```diff
@@ -916,4 +977,20 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
 
     return _fundingCycle.configuration;
   }
+
+  modifier validRates(JBFundingCycleMetadata calldata _metadata) {
```
mejango, line 981: nice

contracts/JBETHPaymentTerminal.sol
```diff
@@ -295,7 +292,7 @@ contract JBETHPaymentTerminal is
         _fundingCycle,
         _distributedAmount,
         _projectOwner,
-        string(bytes.concat('Fee from @', _handle))
+        string(bytes.concat('Fee from ', bytes32(_projectId)))
```
mejango, line 295: Fee from project with ID 1234.  Let's make sure this print out as an int, not gibberish.

contracts/JBETHPaymentTerminal.sol
```diff
@@ -304,7 +301,7 @@ contract JBETHPaymentTerminal is
       _projectId,
       _fundingCycle,
       _distributedAmount - _feeAmount,
-      string(bytes.concat('Payout from @', _handle))
+      string(bytes.concat('Payout from ', bytes32(_projectId)))
```
mejango, line 304: same as above

contracts/JBETHPaymentTerminal.sol
```diff
@@ -360,21 +357,15 @@ contract JBETHPaymentTerminal is
...
 
-    // Get a reference to the handle of the project paying the fee and sending payouts.
-    bytes32 _handle = projects.handleOf(_projectId);
-
     // Take a fee from the _withdrawnAmount, if needed.
     // The project's owner will be the beneficiary of the resulting minted tokens from platform project.
     // The platform project's ID is 1.
+
+    // Prevents stack-too-deep error
+    string memory str = string(bytes.concat('Fee from ', bytes32(_projectId)));
```
mejango, line 365: same as above

contracts/JBETHPaymentTerminal.sol
mejango: @peripheralist getting your eyes here. Because we're removing handles from JBProjects, the memos will only show projectIds. The UI should still be able to resolve these to ENS handles. Wonder if there's a standard way you want to signal to the UI to look resolve a projectId into an ens. maybe with an `@` ? i.e. "Fee from @1234" means => lookup project with ID 1234, get the owner address, try to resolve ENS to dispaly.

mejango (changes requested):
nice.

I'd like @peripheralist's feedback on UI changes from using projectIds in memos. And lets double check that the memos print projectIds as ints and not gibberish bytes.

contracts/JBETHPaymentTerminal.sol
EzraWeller: Yeah, I wasn't sure what to do here, and I don't think what I did is great (Ids will probably come out like 0x1ffffff something ugly like that). TBH, I'm not sure what the memos are for, either, would it be a problem to remove them?

contracts/JBController.sol
```diff
@@ -259,8 +261,11 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
     @dev
     Anyone can deploy a project on an owner's behalf.
 
+    @dev
```
EzraWeller, line 264: Ah, oops, this @dev comment is a leftover, should be removed.

contracts/JBETHPaymentTerminal.sol
mejango: also cc @odd-amphora. reminds me of the wikitoken memo whoopsies :)

mejango: @EzraWeller they're great for UI's to show activity: 
image

we can resolve these in the ENS scheme the same way we resolve names for payouts.

mejango: @EzraWeller take a look at https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b3b83b558ebb9982e27ae5ee0bb5f33f278863dd/contracts/utils/Strings.sol#L15-L35. might be what we need to convert uint256 to string for concating in the string.

EzraWeller: Yeah, was hoping to avoid having to add that extra strings lib -- would it be terrible to just have the front end figure these memos out based on the events? That seems reasonable to me and fits the general philosophy of minimizing on-chain activity, but I guess this is where we need opinions from the UI folks.

Happy to add the strings lib if yall want to.

mejango: yeah, need advice from front end folks.

if front end can parse the projectId from the bytes, great.

peripheralist: the UI will need to do some string interpolation that will get applied to every memo—even non-fee messages—so `@` may be too generic. but i like the idea of using a standard string that other folks could take advantage of if they wanna reference projects in a memo. maybe just... `@@`?

peripheralist: alternatively if we only care about project handle lookups in these memos specifically, the UI could just search for the whole string "Fee from..." "Payout from...", and we don't need a symbol at all.

mejango: i like the standardization so projects can link to other projects from memos! how bout "jb#1234"

EzraWeller: I'm wondering why we need the memos at all. I assume we're getting those by watching for events or using a subgraph -- can we just construct a string like "Fee from bla to bla" using the event, without the memo?

mejango: We could. the way it's currently done is just standardized like all other memos that people can send along with their contributions.

mejango: i.e. these are all send the `_pay` as memos
image

mejango: I suppose there's not much about the events other than the memo that would give us a special signal that they were issued as a `fee` or `payout`.

mejango: I have a feeling the bytes version you have up will work fine, we just need to confirm. can we write a simple test to check? cc @dhyon @drgorillamd for extra eyes on this convo

drgorillamd: The string(bytes.concat('Fee from ', bytes32(_projectId))) is NOT working as intended (uint to bytes32 is ok I guess, but it's then casted to string -> the numeric value ends up as the corresponding ascii (for instance 125 is }  )

uint256 to string should be, as @mejango mentioned handled by OZ lib (or something like https://stackoverflow.com/a/65707309/10974969 )

EzraWeller: Okay, will add something like this.

EzraWeller: Updated, hopefully this works better @drgorillamd

contracts/JBController.sol
dhyon: I like the gist of the idea but does it make sense to increase contract size by inlining three functions with this modifier when we can just throw these checks in the _configure private function that they all call anyways?

mejango: true. Although the checks should come before calling to createProjects imo. Validation before any external fns.

EzraWeller: IMO it would be fine to move these into _configure, since that is the only thing this is validating for, right?

EzraWeller: Did this, lmk if you want it changed back

mejango: perfect, im on board

EzraWeller:
- Now includes #26 (voting capability for NFTs)
- Merged in latest from main
- Attempted to do the changes @mejango et al discussed in discord regarding using blank memos in fees + distributions

Remaining issues:
- Did I do the memo thing right? Don't have my brain 100% wrapped around that
- `ERC721Checkpointable.sol` does not have tests -- maybe we can grab them from a NounsDAO repo? Don't want to or probably have time to write new ones myself atm
- Missing coverage for `JBTerminalUtility.sol` -- I think this will be fixed by #71

contracts/JBController.sol
```diff
@@ -370,25 +371,93 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
...
+    uint256 projectId,
+    JBFundingCycleData calldata _data,
+    JBFundingCycleMetadata calldata _metadata,
+    uint256 _mustStartAtOrAfter,
+    JBGroupedSplits[] memory _groupedSplits,
+    JBFundAccessConstraints[] memory _fundAccessConstraints,
+    IJBTerminal[] memory _terminals
+  ) external returns (uint256 configuration) {
+    // Check the project exists and is owned by msg.sender
+    if (msg.sender != projects.ownerOf(projectId)) revert CALLER_NOT_PROJECT_OWNER();
```
mejango, line 450: this can use a `requiresPermission` modifier. see `reconfigureFundingCyclesFor` tx

contracts/JBController.sol
```diff
@@ -370,25 +371,93 @@ contract JBController is IJBController, JBTerminalUtility, JBOperatable, Reentra
...
+
+    // If there is a previous configuration, reconfigureFundingCyclesOf should be called instead
+    if (fundingCycleStore.latestConfigurationOf(projectId) != 0) {
+      revert PROJECT_ALREADY_LAUNCHED();
```
mejango, line 454: "FUNDING_CYCLES_ALREADY_LAUNCHED"?

contracts/JBETHPaymentTerminal.sol
```diff
@@ -699,8 +676,8 @@ contract JBETHPaymentTerminal is
     if (feeAmount == 0) return 0;
 
     _fundingCycle.shouldHoldFees()
-      ? _heldFeesOf[_projectId].push(JBFee(_amount, uint8(fee), _beneficiary, _memo))
-      : _takeFee(feeAmount, _beneficiary, _memo); // Take the fee.
+      ? _heldFeesOf[_projectId].push(JBFee(_amount, uint8(fee), _beneficiary)) // Take the fee.
```
mejango, line 679: i think the comment got repositioned

mejango (changes requested):
love this.

Few small notes to address.

mejango (approved):
Great!

PR #70, opened 2022-01-23
Splits stored as packed ints instead of structs

mejango:
Splits structs take up two 32byte storage slots. The structs being saved rarely contain an `allocator` or a `lockedUntil` timestamp however. As such, it doesn't make sense to explicitly store these 0 values with each split.

This PR changes the storage method to storing two separate uint256 storage slots, one of which will only be used if an `allocator` or `lockedUntil` timestamp are included in the Split.

We will then also construct the array manually from the two mappings as to not store two separate length values.

JBSplits are also now sent with uint256 values instead of cropped amounts.

No need to check for projectId or lockedUntil overflow since projectId permissions is checked for, and any larger lockedUntil will still be sufficiently far into the future.


contracts/JBSplitsStore.sol
```diff
@@ -183,10 +208,77 @@ contract JBSplitsStore is IJBSplitsStore, JBOperatable {
         revert INVALID_TOTAL_PERCENT();
       }
 
-      // Push the new split into the project's list of splits.
-      _splitsOf[_projectId][_domain][_group].push(_splits[_i]);
+      uint256 _packedSplitParts1 = _splits[_i].preferClaimed ? 1 : 0;
+      _packedSplitParts1 |= _splits[_i].percent << 1;
```
dhyon, line 212: we could probably use an upper bounds check of an individual split's percent above

contracts/JBSplitsStore.sol
```diff
@@ -183,10 +208,77 @@ contract JBSplitsStore is IJBSplitsStore, JBOperatable {
...
+      _packedSplitParts1 |= _splits[_i].projectId << 33;
```
dhyon, line 213: an upper bounds check of the projectId is also good to have

dhyon: max split percentage is 1000000000 so that would be 30 bits; shouldn't we shift by 31 here?

contracts/JBSplitsStore.sol
```diff
@@ -183,10 +208,77 @@ contract JBSplitsStore is IJBSplitsStore, JBOperatable {
...
+      _packedSplitParts1 |= uint256(uint160(address(_splits[_i].beneficiary))) << 89;
+
+      // Store the first spit part.
+      _packedSplitParts1Of[_projectId][_domain][_group][_i] = _packedSplitParts1;
+
+      // If there's data to store in the second packed split part, pack and store.
+      if (_splits[_i].lockedUntil > 0 || _splits[_i].allocator != IJBSplitAllocator(address(0))) {
+        uint256 _packedSplitParts2 = uint48(_splits[_i].lockedUntil);
```
dhyon, line 221: upper bounds check on lockedUntil?

contracts/JBSplitsStore.sol
mejango: true. we originally had it in uint32, which is a cap of 4,294,967,296. I reduced it to 1,000,000,000 but didnt realize this would fit in fewer bits. 

Doesn't hurt to keep it in 32, we dont need the extra bits.

mejango: upper bound check on projectId is good

contracts/JBSplitsStore.sol
mejango: will do

contracts/JBSplitsStore.sol
mejango: isnt the check for the total percents enough?

dhyon: true dat

dhyon (approved):
tremendous

PR #75, opened 2022-01-23
Feature/jbprojects flexible uri

drgorillamd:
- Add IJBTokenUriResolver interface
- Implement tokenURI(projectId) from ERC721Metadata in JBProject


mejango (approved):
yes!

PR #86, opened 2022-01-25
touchups

mejango:


contracts/JBProjects.sol
```diff
@@ -38,27 +37,21 @@ contract JBProjects is ERC721Votes, Ownable, IJBProjects, JBOperatable {
...
+    _projectId The ID of the project to which the metadata belongs.
     _domain The domain within which the metadata applies.
   */
   mapping(uint256 => mapping(uint256 => string)) public override metadataContentOf;
 
+  /**
+    @notice
+    The contract resolving each project ID to its ERC721 URI.
+  */
+  IJBTokenUriResolver public override tokenUriResolver;
```
odd-amphora, line 53: For consistency should we s/Uri/URI everywhere (contract name, local variable, etc.)?

contracts/JBProjects.sol
```diff
@@ -140,18 +133,17 @@ contract JBProjects is ERC721Votes, Ownable, IJBProjects, JBOperatable {
   */
   function tokenURI(uint256 _projectId) public view override returns (string memory) {
```
odd-amphora, line 134: ^ here we use all caps (my preference)

contracts/JBProjects.sol
```diff
@@ -140,18 +133,17 @@ contract JBProjects is ERC721Votes, Ownable, IJBProjects, JBOperatable {
...
     @notice 
-    Set the address of the IJBTokenUriResolver used to retrieve the tokenURI of projects
+    Set the address of the IJBTokenUriResolver used to retrieve the tokenURI of projects.
 
     @param _newResolver The address of the new resolver.
   */
   function setTokenUriResolver(IJBTokenUriResolver _newResolver) external override onlyOwner {
     tokenUriResolver = _newResolver;
-    emit SetTokenUriResolver(_newResolver);
+    emit SetTokenUriResolver(_newResolver, msg.sender);
```
odd-amphora, line 147: Same with event and this fn name.

contracts/JBProjects.sol
mejango: this is overriden and part of the base contract. everywhere else we use Uri

contracts/JBProjects.sol
mejango: this is the standard. we cant change the overriden tokenURI one though. The rest of the JB ecosystem prefers the camel case version.

should we keep consistent in the contract, or in the JB ecosystem?

odd-amphora: Ah yes you're right, sorry forgot about that. camelCaseSoundsGoodToMe

odd-amphora (approved):

PR #111, opened 2022-02-13
fix: critical - recordUsedAllowanceOf

drgorillamd:
Fix bug where projectOwner could use its overflow allowance to take both overflow and target funds.

Solution: compare the used overflow allowance to the remaining overflow (instead of comparing to the whole project balance).

### DIFF
--- a/contracts/JBETHPaymentTerminalStore.sol
+++ b/contracts/JBETHPaymentTerminalStore.sol
@@ -364,21 +364,6 @@ contract JBETHPaymentTerminalStore {
       revert FUNDING_CYCLE_DISTRIBUTION_PAUSED();
     }
 
-    // The new total amount that has been distributed during this funding cycle.
-    uint256 _newUsedDistributionLimitOf = usedDistributionLimitOf[_projectId][fundingCycle.number] +
-      _amount;
-
-    // Amount must be within what is still distributable.
-    uint256 _distributionLimitOf = directory.controllerOf(_projectId).distributionLimitOf(
-        _projectId,
-        fundingCycle.configuration,
-        terminal
-      );
-
-    if (_newUsedDistributionLimitOf > _distributionLimitOf || _distributionLimitOf == 0) {
-      revert DISTRIBUTION_AMOUNT_LIMIT_REACHED();
-    }
-
     // Make sure the currencies match.
     if (
       _currency !=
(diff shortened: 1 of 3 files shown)


contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -476,8 +476,16 @@ contract JBETHPaymentTerminalStore {
...
+    // Get the current funding target
+    uint256 distributionLimit =
+      directory.controllerOf(_projectId).distributionLimitOf(
+        _projectId,
+        fundingCycle.configuration,
+        terminal
+      );
+
+    // The amount being withdrawn must be available in the overflow.
+    if (withdrawnAmount > balanceOf[_projectId] + usedDistributionLimitOf[_projectId][fundingCycle.number] - distributionLimit) {
```
dhyon, line 488: nice catch!

I'm wondering if we should also account for the overflow that can be spread out across multiple terminals like it's done here: https://github.com/jbx-protocol/juice-contracts-v2/blob/e0722447d79de0af79017e6057065bcf9f845ee9/contracts/JBETHPaymentTerminalStore.sol#L677-L681

@mejango thoughts?

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -475,9 +475,35 @@ contract JBETHPaymentTerminalStore {
...
+      directory.controllerOf(_projectId).distributionLimitOf(
+        _projectId,
+        fundingCycle.configuration,
+        terminal
+      );
+    
+    // The current overflow is the balance minus what still needs to be distributed
+    uint256 _leftToDistribute = distributionLimit - usedDistributionLimitOf[_projectId][fundingCycle.number];
+
+    // Get the distribution limit currency (which might or might not be the same as the overflow allowance)
```
drgorillamd, line 490: @mejango no stack issue (yet :D)

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -475,9 +475,35 @@ contract JBETHPaymentTerminalStore {
...
+      ? _leftToDistribute
+      : PRBMathUD60x18.div(
+        _leftToDistribute,
+        prices.priceFor(_distributionLimitCurrency, JBCurrencies.ETH)
+      );
 
-    // The amount being withdrawn must be available.
-    if (withdrawnAmount > balanceOf[_projectId]) {
+    // The amount being withdrawn must be available in the overflow.
+    if (balanceOf[_projectId] == 0 || withdrawnAmount > balanceOf[_projectId] - _leftToDistribute) {
```
drgorillamd, line 506: without testing 0 balance, if 0 balance -> revert with underflow

mejango (approved):
Looks great

PR #112, opened 2022-02-13
fix: low - assertion reorder + mock init in tests

drgorillamd:
This PR fix the order of the tests in JBETHPaymentTerminalStore where in both recordUsedAllowance and recordDistributionFor the currency was assessed before testing the related amounts (ie resp the new allowance left and the new distribution left) 
-> the currency for a 0 allowance or distributionLimit being 0 (and ETH being the currency number 1), this lead to reverting with CURRENCY_MISMATCH() (instead of DISTRIBUTION_AMOUNT_LIMIT_REACHED() or INADEQUATE_CONTROLLER_ALLOWANCE()) when using the 0 allowance or distribution.

Fix: testing for those amounts before checking the currency

### DIFF
--- a/contracts/JBETHPaymentTerminalStore.sol
+++ b/contracts/JBETHPaymentTerminalStore.sol
@@ -473,25 +473,9 @@ contract JBETHPaymentTerminalStore {
     withdrawnAmount = (_currency == JBCurrencies.ETH)
       ? _amount
       : PRBMathUD60x18.div(_amount, prices.priceFor(_currency, JBCurrencies.ETH));
-    
-    // Get the current funding target
-    uint256 distributionLimit =
-      directory.controllerOf(_projectId).distributionLimitOf(
-        _projectId,
-        fundingCycle.configuration,
-        terminal
-      );
-
-    // Convert the target into wei, if needed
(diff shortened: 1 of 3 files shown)


contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -442,42 +441,57 @@ contract JBETHPaymentTerminalStore {
...
+    // Get the current funding target
+    uint256 distributionLimit =
+      directory.controllerOf(_projectId).distributionLimitOf(
+        _projectId,
+        fundingCycle.configuration,
+        terminal
+      );
+
+    // Convert the target into wei, if needed
+    distributionLimit = _currency == JBCurrencies.ETH
```
mejango, line 486: _currency here is the `overflowAllowanceCurrencyOf`. We need `distributionLimitCurrencyOf`.

Hopefully stack doesn't get too deep :|

mejango (commented):
boog

contracts/JBETHPaymentTerminalStore.sol
mejango: (yes, the currency for distributionLimit and overflowAllowance can be different)

mejango (approved):
great

PR #114, opened 2022-02-16
feat: system test: flow launch project -> use allowance -> redeem

drgorillamd:


mejango (commented):
lgtm. thoughts on increasing the MAX_FEE fidelity? rn fee must be in increments of 0.5%. I don't think we're loosing anything if we increase the MAX to the same as the FEE_DISCOUNT max or something. watchu think?

PR #117, opened 2022-02-21
added 3 and 7 day buffer ballots

mejango:
- added 3 and 7 day buffer ballots
- stateOf takes a projectId


drgorillamd (approved):
We should do a system-test flow with them:)

PR #118, opened 2022-02-21
Feat: Audit/ps1

drgorillamd:


mejango (approved):
perfect

contracts/libraries/JBOperations.sol
```diff
@@ -7,7 +7,6 @@ library JBOperations {
   uint256 public constant MIGRATE_CONTROLLER = 3;
   uint256 public constant MIGRATE_TERMINAL = 4;
   uint256 public constant PROCESS_FEES = 5;
-  uint256 public constant SET_HANDLE = 6;
```
mejango, line 10: @drgorillamd should we leave 6 out of the sequence or collapse the rest lol

PR #122, opened 2022-02-23
fix: Reentrancy guards and project-to-terminal checks

dhyon:
- added reentrancy guards to terminal pay and addToBalance methods to address audit issue PVE005
- added project-to-terminal checks in terminal pay and addToBalance methods to prevent situations where callers can mistakenly call the wrong terminal + tests


contracts/JBETHPaymentTerminal.sol
```diff
@@ -246,7 +247,12 @@ contract JBETHPaymentTerminal is
     bool _preferClaimedTokens,
     string calldata _memo,
     bytes calldata _delegateMetadata
-  ) external payable override {
+  ) external payable override nonReentrant {
+    // Verify that this terminal is a terminal of provided project ID
+    if (!directory.isTerminalOf(_projectId, this)) {
```
mejango, line 252: you can abstract this to an `requireTerminalMatch` modifier and reuse in addToBalance, whatchu think?

dhyon: sure that's DRYer

drgorillamd:
noice!

mejango (approved):
awesome

PR #126, opened 2022-02-25
JBProject => JBProjectPayer

mejango:
- changed the contract name to be more descriptive of what it actually does.
- consolidated some internal logic.
- add delegateMetadata param
- add ability to pay arbitrary projectId's from this contract, while setting e default one that gets paid by the fallback fn.


contracts/abstract/JBProjectPayer.sol
```diff
@@ -10,82 +10,90 @@ import './../libraries/JBTokens.sol';
...
     directory = _directory;
+    defaultProjectId = _defaultProjectId;
   }
 
   /** 
     Received funds go straight to the project.
   */
   receive() external payable {
-    _pay(msg.sender, '', false, JBTokens.ETH);
+    if (defaultProjectId > 0)
```
drgorillamd, line 38: shouldn't we else revert() ? Worried of stuck eth

test/jb_project_payer/pay.test.js
```diff
@@ -46,21 +48,21 @@ describe('JBProject::pay(...)', function () {
...
     await expect(
-      jbFakeProject.pay(BENEFICIARY, MEMO, PREFER_CLAIMED_TOKENS, TOKEN, {
+      jbFakeProject.pay(PROJECT_ID_2, BENEFICIARY, MEMO, PREFER_CLAIMED_TOKENS, TOKEN, DELEGATE_METADATA, {
         value: AMOUNT,
       }),
     ).to.not.be.reverted;
   });
 
-  it(`Fallback function should pay funds towards project`, async function () {
+  it(`Fallback function should pay funds towards default project`, async function () {
```
drgorillamd, line 65: Can't pay via the fallback if default project id not set?

contracts/abstract/JBProjectPayer.sol
```diff
@@ -10,82 +10,90 @@ import './../libraries/JBTokens.sol';
...
+      _pay(defaultProjectId, msg.value, msg.sender, '', false, JBTokens.ETH, bytes(''));
   }
 
   /** 
     @notice Allows the project that is being managed to be set.
     @param _projectId The ID of the project that is being managed.
   */
-  function setProjectId(uint256 _projectId) external onlyOwner {
-    projectId = _projectId;
+  function setDefaultProjectId(uint256 _projectId) external onlyOwner {
```
drgorillamd, line 46: A wrong projectId will only get caught later, with TERMINAL_NOT_FOUND(), should we test if before (ie here)?

contracts/abstract/JBProjectPayer.sol
mejango: yes. great.

contracts/abstract/JBProjectPayer.sol
mejango: yes. great.

mejango: im cool with the failure later on. less checks

drgorillamd (approved):
lgtm!

PR #131, opened 2022-02-27
fix: low: underflow instead of error

drgorillamd:
Fix the flow:
- no target set
- balance < allowance
- try to use an allowance > balance

Expected: revert with INADEQUATE_WITHDRAW_AMOUNT

Observed: revert on underflow (balance-withdraw) or on INADEQUATE_PAYMENT_TERMINAL_STORE_BALANCE if a target is set

Fix: refactor (test for amount > balance)

https://discord.com/channels/775859454780244028/911325354218631208/946702526537949204


mejango (approved):
lgtm

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -474,6 +474,11 @@ contract JBETHPaymentTerminalStore {
       ? _amount
       : PRBMathUD60x18.div(_amount, prices.priceFor(_currency, JBCurrencies.ETH));
 
+    // The amount being withdrawn must be at least as much as was expected.
+    if (_minReturnedWei > withdrawnAmount || balanceOf[_projectId] < withdrawnAmount) {
+      revert INADEQUATE_WITHDRAW_AMOUNT();
```
dhyon, line 479: hmm I think this error message only applies to the `_minReturnedWei > withdrawnAmount`  case

I think `INADEQUATE_PAYMENT_TERMINAL_STORE_BALANCE()` is more appropriate for the new case you've added

contracts/JBETHPaymentTerminalStore.sol
```diff
@@ -859,4 +859,4 @@ contract JBETHPaymentTerminalStore {
         ? 0
         : _ethBalanceOf - _ethDistributionLimitRemaining;
   }
-}
+}
```
dhyon, line 862: newline plox

dhyon (commented):
system tests coming through!

contracts/JBETHPaymentTerminalStore.sol
drgorillamd: Is it the withdrawAmount or the terminal balance fault :D ?

mejango: maybe two separate checks?

drgorillamd: done!

dhyon: ah just saw this, thanks for handling it

PR #132, opened 2022-02-27
feat: increase the fee precision

drgorillamd:
Currently, fee are out of 200 (defined as MAX_FEE in JBConstant) -> increase the precision by changing them out of 1 000 000 000, matching the MAX_FEE_DISCOUNT

Test updated + fee now packed in uint32 in JBFee[] -> no change in storage layout, 2 words:
uint256 -> 32b
uint32 -> 4b
address -> 20b

### DIFF
--- a/contracts/JBETHPaymentTerminal.sol
+++ b/contracts/JBETHPaymentTerminal.sol
@@ -123,9 +123,9 @@ contract JBETHPaymentTerminal is
     The platform fee percent.
 
     @dev
-    Out of MAX_FEE.
+    Out of MAX_FEE (50_000_000 / 1_000_000_000)
   */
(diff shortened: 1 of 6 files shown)


contracts/JBETHPaymentTerminal.sol
```diff
@@ -70,7 +70,7 @@ contract JBETHPaymentTerminal is
     @notice
     Maximum fee that can be set for a funding cycle configuration.
   */
-  uint256 private constant _FEE_CAP = 10;
+  uint256 private constant _FEE_CAP = 50_000_000;
```
dhyon, line 73: TIL underscores in numbers 😮

dhyon (approved):

PR #133, opened 2022-02-28
QOL tweaks for system test setup

dhyon:
- main change is that we should be using the ds-test file in our lib folder rather than our own copy-pasta'd version


drgorillamd:
Shouldn't we seize the opportunity to label addresses (there is a new cheatcode for this;)?

dhyon:
> Shouldn't we seize the opportunity to label addresses (there is a new cheatcode for this;)?

sure I'm open to any improvements. could you explain more?

README.md
```diff
@@ -31,11 +31,18 @@ forge build
...
 
 ```bash
-forge test
+forge test -vvv
+```
+
+4. Update Forge periodically:
+
+```bash
+forge update
```
drgorillamd, line 40: foundryup rather ?

README.md
```diff
@@ -31,11 +31,18 @@ forge build
 3. Run tests:
 
 ```bash
-forge test
+forge test -vvv
```
drgorillamd, line 34: add `forge install ds-test` if we want to use the test.sol from lib/

drgorillamd:
> > Shouldn't we seize the opportunity to label addresses (there is a new cheatcode for this;)?
> 
> sure I'm open to any improvements. could you explain more?

using `function label(address, string calldata) external;`, we could label all the contract/multisig by a custom label (new cheatcode in hevm)

README.md
dhyon: gotcha I'll add `git submodule update --init` which is a bit more generic in case we add more submodules later down the line

PR #138, opened 2022-03-03
[WIP] Support ERC-20 terminals

mejango:
This is experimental, but worth checking out:

Things to note:

- `JBETHPaymentTerminalStore` has been generalized to `JBPaymentTerminalStore`. This can be reused (with new deployment) by terminals of any token, including the already written `JBETHPaymentTerminal`.

- the `pay` fn in the `IJBTerminal` spec now takes an `_amount` param. This is ignored in the `JBETHPaymentTerminal` in lieu of msg.value.

- `IJBTerminals` are now required to report the JBCurrency index that it uses to resolve a price feed. This allows us to make the terminal store generic (before, the ETH store was referenced JBCurrencies.ETH)

- The `_totalOverflowDuring` function in the store was also made generic, which caused a modification in the `IJBTerminal` spec from requiring a reporting of `ethBalanceOf` to just `balanceOf`. This turns out to be a great change since it prevents the need for non-ETH terminals to rely on `JBPrices` contract to fulfill it's requirements..


contracts/JBPaymentTerminal.sol
```diff
@@ -0,0 +1,987 @@
...
+  //*********************************************************************//
+  // ---------------------- external transactions ---------------------- //
+  //*********************************************************************//
+
+  /**
+    @notice
+    Contribute ETH to a project.
+
+    @dev
+    The msg.value is the amount of the contribution in wei. The _amount param is ignored.
```
drgorillamd, line 266: nit>

contracts/JBPaymentTerminal.sol
```diff
@@ -0,0 +1,987 @@
...
+    (JBFundingCycle memory _fundingCycle, uint256 _withdrawnAmount) = store.recordUsedAllowanceOf(
+      _projectId,
+      _amount,
+      _currency,
+      _minReturnedWei
+    );
+
+    // Get a reference to the project owner, which will receive tokens from paying the platform fee
+    // and receive any extra distributable funds not allocated to payout splits.
+    address payable _projectOwner = payable(projects.ownerOf(_projectId));
```
drgorillamd, line 430: payable is not needed anymore

contracts/JBPaymentTerminal.sol
```diff
@@ -0,0 +1,987 @@
...
+import './libraries/JBTokens.sol';
+
+import './JBPaymentTerminalStore.sol';
+
+// Inheritance
+import './interfaces/IJBPaymentTerminal.sol';
+import './interfaces/IJBTerminal.sol';
+import './abstract/JBOperatable.sol';
+import '@openzeppelin/contracts/access/Ownable.sol';
+import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
```
drgorillamd, line 21: Reentrancy is not a risk here (no eth and project wanting to use erc777 with a onReceive hook for instance would use another terminal)

mejango: JBPaymentTerminal is abstract and used by both the ETH terminal and the ERC ones

drgorillamd: neeat

contracts/JBPaymentTerminal.sol
mejango: true. seems like we haven't needed it in a while

mejango:
lol

contracts/JBETHPaymentTerminal.sol
```diff
@@ -3,926 +3,54 @@
...
-  function setFee(uint256 _fee) external override onlyOwner {
-    // The max fee is 5%.
-    if (_fee > _FEE_CAP) {
-      revert FEE_TOO_HIGH();
+  function _transferFrom(
+    address _from,
+    address payable _to,
+    uint256 _amount
+  ) internal override {
+    if (_from != address(this)) {
```
drgorillamd, line 47: Since _from is only there for interface implementation, do we really need this test?

mejango: we don't. it's guarenteed that if Token == ETH then this wont be called with a _from that's not equal this.
good catch.

PR #139, opened 2022-03-04
Generic price feeds

mejango:
Decouple JBPrices from chainlink oracles. 

- Allows us to evolve our usage of price feeds over time more easily.
- Allows us to use different providers, like Uniswap TWAPs.
- Allows us to roll multi-hop feeds to resolve the price relationship between two currencies using common base pairs. i.e. through discovering the price of JBX <==> ETH and OHM <==> ETH we can create a price feed for JBX <==> OHM despite there potentially not being one directly from the oracle.


mejango:
For some reason the remote tests are failing on a test that got removed in the PR. passes locally.

test/jb_prices/add_feed_for.test.js
```diff
@@ -1,22 +1,23 @@
 import { expect } from 'chai';
 import { ethers } from 'hardhat';
-import { compilerOutput } from '@chainlink/contracts/abi/v0.6/AggregatorV3Interface.json';
+// import { compilerOutput } from '@chainlink/contracts/abi/v0.6/AggregatorV3Interface.json';
```
drgorillamd, line 3: remove?

test/jb_prices/price_for.test.js
```diff
@@ -1,15 +1,16 @@
 import { expect } from 'chai';
 import { ethers } from 'hardhat';
-import { compilerOutput } from '@chainlink/contracts/abi/v0.6/AggregatorV3Interface.json';
+// import { compilerOutput } from '@chainlink/contracts/abi/v0.6/AggregatorV3Interface.json';
```
drgorillamd, line 3: remove

test/jb_prices/price_for.test.js
```diff
@@ -1,15 +1,16 @@
...
+import jbChainlinkPriceFeed from '../../artifacts/contracts/JBChainlinkPriceFeed.sol/JBChainlinkPriceFeed.json';
 import { deployMockContract } from '@ethereum-waffle/mock-contract';
 import { BigNumber } from '@ethersproject/bignumber';
 import errors from '../helpers/errors.json';
 
-describe('JBPrices::priceFor(...)', function () {
+describe.only('JBPrices::priceFor(...)', function () {
```
drgorillamd, line 9: .only ;)

drgorillamd (approved):

contracts/JBPrices.sol
```diff
@@ -109,12 +97,10 @@ contract JBPrices is IJBPrices, Ownable {
   function addFeedFor(
     uint256 _currency,
     uint256 _base,
-    AggregatorV3Interface _feed
+    IJBPriceFeed _feed
   ) external override onlyOwner {
     // There can't already be a feed for the specified currency.
```
drgorillamd, line 102: Since we know have multiple "oracle provider", wouldn't it be cool to be able to change the feed for a currency (ie if we transition from uniV3 to chainlink for instance)?

mejango: would be. though its another layer of trust that might instead merit a new currency index despite reporting rates for the same currency. meh?

mejango: lingering thoughts here @drgorillamd ?

drgorillamd: Then new terminal/migration/etc, while we are "just" changing oracle? 

But indeed, the safer/trustlesser the better, you're right

mejango: no need for migration, just new `currency` index in JBFundAccessConstraints https://github.com/jbx-protocol/juice-contracts-v2/blob/main/contracts/structs/JBFundAccessConstraints.sol to use a new feed.

PR #140, opened 2022-03-04
erc20 instead of erc20Votes

mejango:
JBToken uses erc20 instead of erc20votes


drgorillamd (approved):

mejango:
tbh still mixed feelings. the tradeoffs here are annoying.

mejango:
doing it. Main reason is that unclaimed tokens aren't onchian gov compatible, which would force projects contributors to claim tokens to participate in governance, incurring big fees along the way.

PR #143, opened 2022-03-09
Optimization/shared store

mejango:
This is turning out to be an important PR. As i began refactoring, lots of old patchy-stuff started unwinding and cancelling each other out:

- the JBPaymentTerminalStore is now shared between any terminal (or address) that calls its functions. It stores information relative to the msg.sender, so various terminals can reuse the storage logic including the JBETHPaymentTerminal and the JBERC20PaymentTerminal.

- Move pay and redeem delegate calls from the store into the terminal.

- cleans up documentation and language to make generic components be understood as generic.

- IJBTerminal no longer needs a `delegate` property as controller.mint and controller.burn are now done directly from the Terminal, not the store. This also removes the need for directory.isTerminalDelegateOf. This is also great since the term `delegate` was overloaded.

- Terminal no longer has to combine 'prefersUnclaimedTokens' and 'beneficiary' into a uint256 arg when calling the store. Only the beneficiary needs to be passed.

- all checks for `minReturnedTokens` are moved into the terminal, removing the need to pass that data into the store.


contracts/JBPaymentTerminal.sol
```diff
@@ -0,0 +1,1089 @@
...
+  address public immutable override token;
+
+  /**
+    @notice
+    The platform fee percent.
+
+    @dev
+    Out of MAX_FEE (50_000_000 / 1_000_000_000)
+  */
+  uint256 public override fee = 50_000_000; // 5%
```
drgorillamd, line 131: 2.5% ?

mejango: yes

drgorillamd (approved):

contracts/JBERC20PaymentTerminal.sol
```diff
@@ -42,7 +42,9 @@ contract JBERC20PaymentTerminal is JBPaymentTerminal {
     address payable _to,
     uint256 _amount
   ) internal override {
-    IERC20(token).transferFrom(_from, _to, _amount);
+    _from == msg.sender ?
```
mejango, line 45: shouldn't this be _from == address(this) ?

drgorillamd: yes, absolutely right!

PR #150, opened 2022-03-12
price feed takes decimals as input

mejango:
Potentially important optimization I came across when refining JBPrices docs:

JBPrices no longer has TARGET_DECIMALS. Instead, each request for a price includes the desired decimal fidelity.

Why:

- remove tight coupling to 18 decimal fixed point amounts when converting amounts between different terminal currencies. Opens the door for ERC721 and ERC1155 terminals, with conversion rates between these assets using custom price feeds. 1 Punk == 100 Doodles? sure. 

TBH there are probably other weird complexities when introducing NFT-based terminals (should NFT with tokenId X really be treated the same as the same NFT with tokenId Y? How do fees work?), but at least the reliance on 18 decimals price feeds wont be one of them.

WIP tests and inline docs.

### DIFF
--- a/contracts/JBPaymentTerminalStore.sol
+++ b/contracts/JBPaymentTerminalStore.sol
@@ -2,7 +2,6 @@
 /* solhint-disable comprehensive-interface*/
 pragma solidity 0.8.6;
 
-import '@paulrberg/contracts/math/PRBMathUD60x18.sol';
 import '@paulrberg/contracts/math/PRBMath.sol';
 
 import './interfaces/IJBPrices.sol';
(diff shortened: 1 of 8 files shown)


contracts/JBPaymentTerminalStore.sol
```diff
@@ -131,19 +130,21 @@ contract JBPaymentTerminalStore {
 
     @param _terminal The terminal for which the overflow is being calculated.
     @param _projectId The ID of the project to get overflow for.
+    @param _decimals TODO
```
drgorillamd, line 133: why not _terminal.decimals()?

drgorillamd: oh, no, discard, can be in another currency

PR #151, opened 2022-03-12
use terminal currency instead of passed in one

mejango:


mejango:
@drgorillamd u right. these should just call terminal.currency() since the local stored balance is in terms of that currency. providing any other currency would be wrong

drgorillamd (approved):

PR #152, opened 2022-03-12
Optimization/ijbprices interface view

mejango:
if anything the decimals journey would have been worth it to have caught this lol


drgorillamd (approved):
ooooops

PR #155, opened 2022-03-12
Optimization/metadata

mejango:
rename _delegateMetadata and also forward to data source.


mejango:
goddamn shouldve landed this before 157. will do it again after.

PR #157, opened 2022-03-13
terminals restricted to 18 decimals

mejango:
This is an important PR:

We introduced an ERC-20 terminal recently, however several common ERC-20 implementations dont use 18 decimals. Due to the prior hard-coded expectation of 18 decimals in `JBPrices`, ERC-20's that don't use this amount would have caused dangerous inconsistencies. 

This also opens the door for future Terminals and Stores that use different decimal configurations.

Changes in this PR:

- rename `JBPaymentTerminal` abstract class to `JB18DecimalPaymentTerminal` and moved into /abstract,  and `IJBPaymentTerminal` to `IJB18DecimalPaymentTerminal`
- rename `JBPaymentTerminalStore` abstract class to `JB18DecimalPaymentTerminalStore`.
- rename `JBERC20PaymentTerminal` to `JB18DecimalERC20PaymentTerminal`.
- rename `IJBTerminal`  to `IJBPaymentTerminal`
- pulls the `TARGET_DECIMAL = 18` constant out of `JBPrices` and into `JB18DecimalPaymentTerminalStore`. Pass it into `prices.priceFor(...)`.
- `JB18DecimalERC20PaymentTerminal` requires its token to use 18 decimals, enforced in the constructor.

All tests for the above changes are up to date with 100% coverage.
The following item still needs tests. They are `skip`ed in the suite... i there's two of them and a few more to get coverage back to 100:

- in `JB18DecimalPaymentTerminalStore._currentTotalOverflow` (previously `_totalOverflowDuring(...)`), make the calculation by simply reading from each terminal's `currentEthOverflowOf()` (new). This implementation removes the need for `balanceOf()` and `remainingDistributionLimitOf()` from `IJBPaymentTerminal` (tests have already been removed alongside), and adds `currentEthOverflowOf()` as mentioned. Adjusts `JB18DecimalPaymentTerminal` accordingly by adding a shared implementation of `currentEthOverflowOf()` (needs tests).

@drgorillamd would you mind getting these tests? might also help grok what i did in this PR.

________

Hate to keep shifting things around last minute, but also glad to be catching things now.


contracts/JB18DecimalPaymentTerminalStore.sol
```diff
@@ -760,68 +784,34 @@ contract JBPaymentTerminalStore {
...
+    uint256 _totalOverflow18Decimal = _currency == JBCurrencies.ETH
+      ? _ethOverflow
+      : PRBMath.mulDiv(_ethOverflow, 10**18, prices.priceFor(JBCurrencies.ETH, _currency, 18));
+
+    // Adjust the decimals of the fixed point number if needed to match the target decimals.
     return
-      _currencyBalanceOf <= _currencyDistributionLimitRemaining
-        ? 0
-        : _currencyBalanceOf - _currencyDistributionLimitRemaining;
+      TARGET_DECIMALS == 18
```
drgorillamd, line 813: Always true (for this particular store at least)

contracts/abstract/JB18DecimalPaymentTerminal.sol
```diff
@@ -171,39 +174,25 @@ abstract contract JBPaymentTerminal is IJBPaymentTerminal, JBOperatable, Ownable
...
-        _projectId,
-        _fundingCycleConfiguration,
-        this
-      ) - store.usedDistributionLimitOf(this, _projectId, _fundingCycleNumber);
+      currency == JBCurrencies.ETH
+        ? _overflow
+        : PRBMath.mulDiv(
+          _overflow,
+          10**store.TARGET_DECIMALS(),
+          store.prices().priceFor(currency, JBCurrencies.ETH, store.TARGET_DECIMALS())
```
drgorillamd, line 194: same call to store.TARGET_DECIMALS() twice

PR #161, opened 2022-03-13
it builds

mejango:
Make Store and PaymentTerminals generic again, maintaining the decimal fixes from #157 

More info to come.

This feels fairly final. Tests need to be adjusted. It might be worth us going through the tests again one-by-one, cleaning up loose variables and odds-and-ends as needed.


contracts/JBPaymentTerminalStore.sol
```diff
@@ -163,16 +169,17 @@ abstract contract JBPaymentTerminalStore is IJBPaymentTerminalStore {
...
   */
-  function currentTotalOverflowOf(uint256 _projectId, uint256 _currency)
-    external
-    view
-    returns (uint256)
-  {
-    return _currentTotalOverflowOf(_projectId, _currency);
+  function currentTotalOverflowOf(
+    uint256 _projectId,
+    uint256 _decimals,
```
drgorillamd, line 179: Shouldn't we, for frontend's sake, store decimal[currency] in JBCurrency?

contracts/structs/JBTokenAmount.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+struct JBTokenAmount {
```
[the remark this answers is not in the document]
mejango: how much do params cost?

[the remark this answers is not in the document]
mejango: this info never gets stored

contracts/JBPaymentTerminalStore.sol
mejango: Not sure what you mean.

contracts/structs/JBTokenAmount.sol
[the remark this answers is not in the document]
drgorillamd: meh, 30 units difference (without all the overhead it would add...) - not worth it idd

contracts/JBPaymentTerminalStore.sol
```diff
@@ -256,31 +254,30 @@ abstract contract JBPaymentTerminalStore is IJBPaymentTerminalStore {
...
     @return fundingCycle The project's funding cycle during which payment was made.
-    @return weight The weight according to which new token supply was minted, as a fixed point number with 18 decimals.
     @return tokenCount The number of project tokens that were minted, as a fixed point number with 18 decimals.
     @return delegate A delegate contract to use for subsequent calls.
     @return memo A memo that should be passed along to the emitted event.
   */
   function recordPaymentFrom(
     address _payer,
-    uint256 _amount,
+    JBTokenAmount memory _amount,
```
drgorillamd, line 271: calldata instead of memory (should be -2k)

drgorillamd: Hmm, I'll go hunt for these actually, we might save additional gas with it

contracts/JBPaymentTerminalStore.sol
drgorillamd: Not sure what I meant, discard ':)

contracts/JBPaymentTerminalStore.sol
mejango: oo im real curious

PR #162, opened 2022-03-14
no fees if jbdao hasn't added a terminal for a token

mejango:
If jbdao hasn't added a terminal for a particular token, don't charge fees. Otherwise, the tx's that charge fees will fail.


drgorillamd (approved):
oh yes, good catch!

mejango:
@drgorillamd can i get unit test help for this one too plz:]

mejango:
I'm thinking we might also wanna remove the projectId = PROTOCOL_PROJECT_ID stipulation for not charging fees. 
In case we extend fees in the future to directly put pressure on an AMM pool, the protocol project should pay fees to the same mechanic.

mejango:
> > I'm thinking we might also wanna remove the projectId = PROTOCOL_PROJECT_ID stipulation for not charging fees. In case we extend fees in the future to directly put pressure on an AMM pool (instead of adding to JuiceboxDAO treasury), the protocol project should pay fees to the same mechanic.
> > Just seems more extensible if all projects are treated the same. And lets us remove some code:]
> > wyt @drgorillamd ?
> 
> Hmm, and still keeping something we can change then (ie a bool `isPlatformExemptedFromFee` for instance)? All in all the same but with some gas saving when using the "vanilla" fee destination (which would, imo, happens more often)

maybe? I'm reflecting a bit on the V1 fee dynamic where JuiceboxDAO pays a fee to itself, its not the end of the world and actually does have a cool factor (juiceboxDAO works exactly like all other projects)

PR #169, opened 2022-03-16
Gas optim/Replace (some) memory by calldata in public function

drgorillamd:
Only a few `memory` can be changed to `calldata`, at least without any other modification (ie only external/public function args, or an internal function only if it's an untouched calldata passing through) + calldata's are pushed into the stack -> lots of stacks are then too deep (will try later to solve it as those are in terminal, gas optim could be nice then)

system-test flow gas gain: 
image

Per functions:
image
image


mejango:
nice. marginal, but very nice. Let's see if its worth the stack-too-deep rearranging, which we've already done a lot of (tough to squeeze more stack space in many places)

mejango:
handled in #177

PR #173, opened 2022-03-20
refactor/check effect interaction

drgorillamd:
- Some (minor) suboptimal check-effect-interaction patterns, reorganised to avoid any untrusted call before further state change
- Reentrancy guard removal in JBController
- ReentrancyGuard moved from JBPayoutRedemptionPaymentTerminal to the corresponding call to JBPaymentTerminalStore (allowing terminal reentrancy into different function, after having updated corresponding store state).
-- only function not having a store equivalent being `processFees`


contracts/JBController.sol
```diff
@@ -708,12 +702,12 @@ contract JBController is IJBController, JBOperatable, ReentrancyGuard {
...
       _distributeReservedTokensOf(_projectId, '');
 
-    // Make sure the new controller is prepped for the migration.
-    _to.prepForMigrationOf(_projectId, this);
-
     // Set the new controller.
     directory.setControllerOf(_projectId, _to);
 
+    // Make sure the new controller is prepped for the migration.
+    _to.prepForMigrationOf(_projectId, this);
```
mejango, line 709: prepForMigration should be called prior to the terminal being prepped being the active controller. 

See line 71. This check is important to make sure any resetting logic doesn't get called while a project has an in-progress tokenTracker.

mejango (changes requested):
awesome. one lil change rec

PR #175, opened 2022-03-22
Docs/jbstore

mejango:
A few changes to the view api. Also changes to funding cycle metadata property for better documentation consistency `shouldUseLocalBalanceForRedemptions` => `useTotalBalanceForRedemptions`

started tests, but gonna make progress on write fn docs instead. @drgorillamd 🙏 

updates on write fns:

- remove `beneficiary` arg from recordPay and recordRedemption, the data source shouldnt be making a decision using this arg. This also allows the params to be calldata instead of memory.


contracts/JBPaymentTerminalStore.sol
```diff
@@ -185,34 +175,49 @@ contract JBPaymentTerminalStore is IJBPaymentTerminalStore, ReentrancyGuard {
...
-        _terminal,
+    // Get a reference to the project's current funding cycle.
+    JBFundingCycle memory _fundingCycle = fundingCycleStore.currentOf(_projectId);
+
+    // Get the amount of current overflow.
+    // Use the project's total overflow across all of its terminals if the flag species specifies so. Otherwise, use the overflow local to the specified terminal.
+    uint256 _currentOverflow = _useTotalOverflow
+      ? _currentTotalOverflowOf(_projectId, _terminal.decimals(), _terminal.currency())
+      : _overflowDuring(
+        IJBPaymentTerminal(msg.sender),
```
drgorillamd, line 210:
```suggestion
        _terminal,
```

contracts/JBPaymentTerminalStore.sol
```diff
@@ -185,34 +175,49 @@ contract JBPaymentTerminalStore is IJBPaymentTerminalStore, ReentrancyGuard {
...
 
     @return The amount of overflowed tokens that can be reclaimed.
   */
-  function reclaimableOverflowOf(
+  function currentReclaimableOverflowOf(
     IJBPaymentTerminal _terminal,
     uint256 _projectId,
-    uint256 _tokenCount
+    uint256 _tokenCount,
+    bool _useTotalOverflow
```
drgorillamd, line 200: why are we not reading it from the funding cycle metadata?

mejango: In this case cus the outside world can view what the reclaimable overflow would be in either case. Otherwise the only way to "know" what the value would be is to reconfigure, or do the calculation manually.

PR #177, opened 2022-03-23
Docs/jbterminal

mejango:


contracts/abstract/JBPayoutRedemptionPaymentTerminal.sol
```diff
@@ -860,15 +882,23 @@ abstract contract JBPayoutRedemptionPaymentTerminal is
...
 
+            // Trigger any inherited pre-transfer logic.
             _beforeTransferTo(address(_terminal), _netPayoutAmount);
 
+            // Get a reference to the destination terminal's decimals.
+            uint256 _decimals = _terminal.decimals();
+
+            // If the destination terminal uses a different number of decimals than this terminal, adjust the sent amount accordingly.
+            if (_decimals != decimals)
+              _amount = JBFixedPointNumber.adjustDecimals(_amount, decimals, _decimals);
```
drgorillamd, line 900: Shouldn't this be
_netPayoutAmount = JBFixedPointNumber.adjustDecimals(_netPayoutAmount, decimals, _decimals);

mejango: pushed up a solution that i think makes sense. The terminal will allow transfer given its set decimals, and set the transfer amount given the receiver's set decimals. its up to the receiving terminal to figure it out.
this is edge edge edge case, any properly setup terminal with the same token will have the same decimals.

PR #191, opened 2022-03-28
adjust fc metadata indexes. check for allowance when setting terminal…

mejango:
…s and controllers.

This prevents a project from changing or removing terminals or controllers on demand, which could prevent further interactions to the project or alter expectations dramatically. 

needs tests.


mejango:
lgtm. @aeolianeth merging this into main. 0.0.16 should be up shortly.

PR #193, opened 2022-03-29
Add to balance if split has no beneficiary

mejango:
when distributing payout splits, `addToBalance(...)` instead of `pay(...)` if the split does not specify a beneficiary.

needs tests in the dreaded `distributePayoutsOf` suite.


mejango:
need to trigger paydelegate. abort.

mejango:
@drgorillamd reopened this (fak), and wrote the tests to get it to 100. If the option is to error on beneficiary == 0 and break distribute, or to addToBalance instead, i feel we might as well add the feature.

drgorillamd (approved):

PR #195, opened 2022-03-30
Fix/coverage

drgorillamd:
Since our update to Solidity coverage 0.8.0beta, some branches were discovered as non-covered - moar tests to cover them

+ only parts non-covered are reentrancy guards


mejango (approved):
lgtm

PR #207, opened 2022-04-02
fix

mejango:
Pretty nasty bug. pretty straight forward.
1. project 1 issueToken tokenA . 
2. project 2 changeToken to tokenA.

Since JBTokenStore owns tokenA from project #1 issuing it, project 2 can now mint/burn tokenA.
 
Fix:
- make sure that if tokenA is attached to project #1, it can't also be attached to project 2. This seems cleanest. Tradeoff is removes same token for multiple projects functionality, but this seems reasonable.


drgorillamd (approved):
good good good catch

drgorillamd:
Aaah, no, it's the != the cheapest (but disappears with Optim anyway)

PR #213, opened 2022-04-06
added return value in pay to improve composability

mejango:
Added return values to `distributePayouts`, `useOverflowAllowance`, `pay`, `migrate`.

Helps with some composable functionality.


drgorillamd (approved):
lgtm! And will make forge-thingies easier

PR #219, opened 2022-04-08
splits payer contract similar to project payer

mejango:
add a direct payment contract that splits between JBSplits


contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -0,0 +1,280 @@
...
+            _split
+          );
+          // Trigger the allocator's `allocate` function.
+          _split.allocator.allocate{value: _settleAmount}(_data);
+          // Otherwise, if a project is specified, make a payment to it.
+        } else if (_split.projectId != 0) {
+          _pay(
+            _split.projectId,
+            _token,
+            _amount,
```
drgorillamd, line 246: _settleAmount instead?

contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -0,0 +1,280 @@
...
+          // Create the data to send to the allocator.
+          JBSplitAllocationData memory _data = JBSplitAllocationData(
+            _settleAmount,
+            _decimals,
+            _PROTOCOL_PROJECT_ID,
+            0,
+            _split
+          );
+          // Trigger the allocator's `allocate` function.
+          _split.allocator.allocate{value: _settleAmount}(_data);
```
drgorillamd, line 240: ``
```suggestion
          _split.allocator.allocate{value: _token == JBTokens.ETH ? _settleAmount : 0}(_data);
```

contracts/JBETHERC20SplitsPayer.sol
drgorillamd, line 246:
```suggestion
            _settleAmount,
```

contracts/JBETHERC20SplitsPayer.sol
drgorillamd: meh, stack too deep then ':)

drgorillamd: Suggest then
```
          // Trigger the allocator's `allocate` function (avoiding stack too deep)
          if(_token == JBTokens.ETH) _split.allocator.allocate{value: _settleAmount }(_data);
          else _split.allocator.allocate(_data);
  ```

File: contracts/JBETHERC20SplitsPayer.sol [hunk 3]
```diff
@@ -0,0 +1,412 @@
...
+    } else {
+      _payer = address(this);
+      _amount = msg.value;
+      _decimals = 18;
+    }
+
+    // Route the payment to the splits.
+    // Pay the split and get a reference to the amount paid.
+    uint256 _leftoverAmount = _payToSplits(
+      defaultProjectId,
```
drgorillamd, line 224: `defaultSplitsProjectId` instead? (defaultProjectId is inherited from project payer)

contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -0,0 +1,412 @@
...
+
+      if (_splitAmount > 0) {
+        // Transfer tokens to the split.
+        // If there's an allocator set, transfer to its `allocate` function.
+        if (_split.allocator != IJBSplitAllocator(address(0))) {
+          // Create the data to send to the allocator.
+          JBSplitAllocationData memory _data = JBSplitAllocationData(
+            _payer,
+            _splitAmount,
+            0,
```
drgorillamd, line 366: Shouldn't we pass the decimal? One allocator can have multiple splitsPayer/multiple token, no?

contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -0,0 +1,412 @@
...
+            defaultProjectId,
```
drgorillamd, line 367: same, `defaultSplitsProjectId` instead?

drgorillamd, line 367: `_splitsProjectId` ?

contracts/JBETHERC20SplitsPayer.sol
mejango: we should def pass the decimal.

contracts/JBETHERC20SplitsPayer.sol
mejango: Leave it as is. The `defaultSplitsProjectId` is just used to index the project to which the splits belong. This should, in most practical cases I can think of, be the same as the `defaultProjectId` inherited. But deployers may wanna customize the payout of leftover some other way.

contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -0,0 +1,412 @@
...
+          18, // decimals.
+          defaultBeneficiary,
+          0, // min returned tokens.
+          defaultPreferClaimedTokens,
+          defaultMemo,
+          defaultMetadata
+        );
+    // If no project was specified, send the funds directly to the beneficiary or the msg.sender.
+    else
+      Address.sendValue(
```
drgorillamd, line 153: sendValue will only work to send eth

mejango: nice. push going up

drgorillamd (approved):

PR #222, opened 2022-04-10
IJBPaymentTerminal no longer opinionated about having 1 token, 1 deci…

mejango:
…mal value, and 1 currency

pure interface optimization. no changes to implementations.

jk slight change in director.setTerminalsOf where the primary terminal isn't removed out of necessity.


contracts/structs/JBFundAccessConstraints.sol
```diff
@@ -6,6 +6,8 @@ import './../interfaces/IJBPaymentTerminal.sol';
 struct JBFundAccessConstraints {
   // The terminal within which the distribution limit and the overflow allowance applies.
   IJBPaymentTerminal terminal;
+  // The token for which the distribution limit applies.
```
drgorillamd, line 9: Nit> "The token for which the fund access constraint applies."?

mejango: nice

drgorillamd (approved):

PR #240, opened 2022-04-22
Forge tests for EIP165

xBA5ED:
- Added new forge tests
- deleted old mocha tests
- moved custom errors into contract scope instead of global scope


drgorillamd:
Foundry is just so f cool isn't it:)?

PR #247, opened 2022-05-02
Smol: change minting access control in controller

drgorillamd:
Allows the current funding cycle's datasource to mint without being part of the authorized domain


mejango (approved):
nice

PR #248, opened 2022-05-03
done

mejango:


deploy/deploy.js
```diff
@@ -211,44 +214,41 @@ module.exports = async ({ deployments, getChainId }) => {
   if ((await jbDirectoryContract.connect(deployer).owner()) != multisigAddress)
     await jbDirectoryContract.connect(deployer).transferOwnership(multisigAddress);
 
-
   // If needed, deploy the protocol project
   if ((await jbProjects.connect(deployer).count()) == 0) {
-
     console.log('Adding reserved token splits with current beneficiaries (as of deployment)');
```
drgorillamd, line 219: Should we keep this in the repo (is some people try to fork without reading the script)?

drgorillamd (approved):
Done

deploy/deploy.js
mejango: good q. I think we should commit exactly what we deploy, and make another "generic" version.

PR #255, opened 2022-05-08
second controller and terminal contracts

mejango:
Created a second iteration of JBController, and second iterations of terminals that use the second iteration of the controller.

let me know what you think of the folder structure. I essentially just made a folder with the filename and renamed the files to `1.sol`and added `2.sol`.


mejango:
merp, just realized this wont work since the call to distribute reserved is still based on the funding cycle reserved rate. 
but the versioning strategy is still interesting in the PR. seems to work fine.

drgorillamd (approved):

PR #259, opened 2022-05-11
Offering NFTs based on contributions to treasuries (JBNFTRewardsDataSourceDelegate.sol, JBToken721, JBToken721Store)

tankbottoms:
* - add an implementation of an ERC721 token

* - WIP ERC721 store

* - 721 store checkpoint

* - project files

* - updated dependencies (works with node 16.14.2)

* - updated dependencies
- remove rari for name collision & copy it in

* - add 721 store to controller

* - remove super

* - NFT reward tier definition

* - NFT uri resolvers

* - opensea-style metadata

* - uri resolvers

* - allow for multiple NFT contracts per project
- allow (de)registstration of compatible external NFT contracts
- remove delayed claims, NFTs will be minted immediately to preserve sequence
- add NFTTranche to funding cycle

* - allow tranches to point to different contracts

* - alias rariERC721

* add rari solmate to foundry remappings

* - Funding cycle NFT rewards hooks

* - remove IJBTokenContractUriResolver

* - add mintcap to nfttranche

* - extra reference

* - readme update for forge

* - opt out of redeem workflow

* - more forge mocks

* - mint via controller, since it deploys the token

* - return token id on mint

* - console for forge tests

* - forge nft workflow test

* - missing permission

* - missing permission

* - permissions for reward nft mint

* rollup

* rollup

* - syntax improvements

* - post-rollup update

* - remove unsed features

* - updated existing tests

* - corrected error message

* - docs

* - remove NFT features that aren't needed immediately

* - controller NFT tests

* - JBToken721Store tests

* - JBToken721 tests

* - more tests

* - more tests

* - pass memo along

* - JBNFTRewardDataSourceDelegate tests

* - increased test coverage

* - docs
- increment bug fix

* - forge test update

* - docs
- more tests
- uri setting functions on token721

* - JBNFTRewardDataSourceDelegate tests

Co-authored-by: dhyon <hyperformance@gmail.com>


README.md
```diff
@@ -43,13 +43,19 @@ curl -L https://foundry.paradigm.xyz | sh
 git submodule update --init
 ```
 
-3. Run tests:
+3. Update foundry to get the baseline install:
```
drgorillamd, line 46: isn't the shell script spinning up the last version already? Haven't checked tbh

contracts/JBController/1.sol
```diff
@@ -478,8 +488,9 @@ contract JBController is IJBController, IJBMigratable, JBOperatable, ERC165 {
     returns (uint256 configuration)
   {
     // If there is a previous configuration, reconfigureFundingCyclesOf should be called instead
-    if (fundingCycleStore.latestConfigurationOf(_projectId) > 0)
+    if (fundingCycleStore.latestConfigurationOf(_projectId) > 0) {
```
drgorillamd, line 491: why?

contracts/JBController/1.sol
```diff
@@ -695,6 +750,54 @@ contract JBController is IJBController, IJBMigratable, JBOperatable, ERC165 {
...
+  */
+  function mintTokens721Of(
+    uint256 _projectId,
+    address _beneficiary,
+    string calldata _memo
+  )
+    external
+    virtual
+    override
+    requirePermissionAllowingOverride(
```
drgorillamd, line 774: Suggest moving this (and not use the modifier) inside the function body, to avoid calling fundingCycleStore twice for the same data

contracts/JBController/1.sol
```diff
@@ -695,6 +750,54 @@ contract JBController is IJBController, IJBMigratable, JBOperatable, ERC165 {
...
+        address(fundingCycleStore.currentOf(_projectId).expandMetadata().dataSource) == msg.sender
+    )
+    returns (uint256 tokenId)
+  {
+    // Scoped section prevents stack too deep. `_fundingCycle` only used within scope.
+    {
+      // Get a reference to the project's current funding cycle.
+      JBFundingCycle memory _fundingCycle = fundingCycleStore.currentOf(_projectId);
+
+      // If the message sender is not a terminal, the current funding cycle must allow minting.
```
drgorillamd, line 788: So no token mint possible == no NFT mint neither, correct?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,133 @@
...
+    returns (
+      uint256 reclaimAmount,
+      string memory memo,
+      IJBRedemptionDelegate delegate
+    )
+  {
+    return (0, '', IJBRedemptionDelegate(address(0)));
+  }
+
+  function didPay(JBDidPayData calldata _data) external override {
```
drgorillamd, line 107: You might want to add an access control to this;)
ie msg.sender == terminal for instance?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,133 @@
...
+    // NOTE: JBToken721Store has onlyController(_projectId) on mintFor which will revert for incorrect sender.
```
drgorillamd, line 108: I think there is still a hole here:
didPay is open -> if called by a random eao:
controller.mintTokens721Of takes DataSource as authorized msg.sender (passes)
JBToken721Store.minFor takes the controller as authorized msg.sender (passes)
-> someone calling didPay will be able to effectively mint

contracts/JBToken721.sol
```diff
@@ -0,0 +1,305 @@
...
+    Mints the next NFT id.
+
+    @dev
+    Only the owner of this contract can mint more of it.
+
+    ignored: _projectId The ID of the project to which the token belongs. This is ignored.
+    @param _account The account to mint the tokens for.
+  */
+  function mint(uint256, address _account) external override onlyOwner returns (uint256) {
+    uint256 tokenId = _nextTokenId;
```
drgorillamd, line 180: quick trick to avoid natspec/solc complaining of unused var, while keeping to prototype (for reusability):

```
    @param _projectId The ID of the project to which the token belongs. This is ignored.
    @param _account The account to mint the tokens for.
  */
  function mint(uint256 _projectId, address _account) external override onlyOwner returns (uint256) {
      _projectId; // Will not change the bytecode, but solc and natspec happiness are now achieved
      uint256 tokenId = _nextTokenId;
```

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,166 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import './helpers/console.sol';
+import './helpers/TestBaseWorkflow.sol';
+import '../structs/JBTokenAmount.sol';
+import '../libraries/JBCurrencies.sol';
+
+interface Vm {
```
drgorillamd, line 9: this is part ot the TestBaseWorkflow (via evm.warp)

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,166 @@
...
+  address private _projectOwner;
+  uint256 private _reservedRate = 5000;
+
+  JBNFTRewardDataSourceDelegate private _jbNFTRewardDataSourceDelegate;
+  IJBToken721 private _nft;
+
+  function setUp() public override {
+    super.setUp();
+
+    Vm vm = Vm(address(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D));
```
drgorillamd, line 36: same, you can already use it via `evm.`

drgorillamd: (I should add the forge-std to follow cheat code addition tho, instead of the hardcoded interface rn)

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,166 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
+
+import './helpers/console.sol';
```
drgorillamd, line 4: should we keep this (+we used emit log until now, but up's to anyone's choice:)?

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,166 @@
...
+    );
+
+    _fundingCycleData = JBFundingCycleData({
+      duration: 60 * 60,
+      weight: 1000 * 10**18,
+      discountRate: 450000001,
+      ballot: IJBFundingCycleBallot(address(0))
+    });
+
+    _fundingCycleMetadata = JBFundingCycleMetadata({
```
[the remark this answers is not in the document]
drgorillamd: is there a reason to use a reconfig? dataSource is not called until the pay anyway

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,166 @@
...
+      _groupedSplits,
+      _fundAccessConstraints,
+      ''
+    );
+  }
+
+  function testNFTRewardMint() public {
+    _terminals[0].pay{value: 2 * 10**18}(
+      _projectId,
+      1 * 10**18,
```
drgorillamd, line 154: amount can be whatever when a msg.value is passed, it is however a bit disturbing when they differ;)
+ I think this could be fuzzed

test/jb_nft_reward_datasource_delegate/did_pay.test.js
```diff
@@ -0,0 +1,279 @@
...
+    };
+  }
+
+  it(`Should mint token if meeting contribution parameters`, async function () {
+    const { timestamp, jbEthPaymentTerminal, addrs } = await setup();
+    const caller = addrs[0];
+    const beneficiary = addrs[1];
+
+    expect(
+      await jbEthPaymentTerminal
```
drgorillamd, line 179: This should test the delegate instead, not the terminal/e2e

contracts/JBController/1.sol
tankbottoms: cool, thanks

contracts/JBToken721.sol
tankbottoms: cool, was wondering why this was done

contracts/system_tests/TestNFTRewardsFlow.sol
tankbottoms: that would be cool, its useful for amateurs like me

README.md
cptspacecadet: instructions in the readme didn't work for me until `foundryup` was run once

contracts/JBController/1.sol
cptspacecadet: can be changed, but the new functions were modeled on existing ones as not to rock the boat too much

cptspacecadet:
> 

The idea is to allow project managers to optionally associate NFT rewards with contributions for every funding cycle. This would support the genesis NFT idea that's being discussed but also allow for other rewards to be created. `JBNFTRewardDataSourceDelegate` can mint NFTs for the first *N* contributors for a project in a given funding cycle who sent more than some amount. It supports ether and token contributions.

mejango:
@cptspacecadet do we need the new JBController or token store then?

tankbottoms:
So make the controller the owner of the token?

drgorillamd:
> * checkpoint

+ the minimal example reproducing this behaviour (data source/delegate being a NFT) here: https://github.com/jbx-protocol/juice-extension-templates/tree/main/contracts/NFT-example

contracts/JBController/1.sol
drgorillamd: > can be changed, but the new functions were modeled on existing ones as not to rock the boat too much

https://github.com/jbx-protocol/juice-contracts-v2/blob/fc7090647d19ca3dee5e648e55ea16284973d8b5/contracts/JBController/2.sol#L645

contracts/system_tests/TestNFTRewardsFlow.sol
drgorillamd: > that would be cool, its useful for amateurs like me

yh yh, I use it all the time too, but `emit log(..)` is part of ds-test already (no need to add a new helper then)

test/jb_nft_reward_datasource_delegate/did_pay.test.js
drgorillamd: + same for all the following tests, no real unit tests

tankbottoms:
> @tankbottoms im a bit confused about "owner of token". In this case, the Delegate/Data source could also be the NFT. As payments come in, you can `_mint` using internal fns.
> 
> see this convo: https://discord.com/channels/775859454780244028/786985806396784690/950560349852143626

ok, ok, you guys want it in the Delegate/Data (dd) source, it just looks super weird to me. from the angle of just building a bunch more extensions and being able to chain them together, i can avoid shoving a bunch of logic in the dd, especially since the token for the purposes of abstracting tokenUriResolver, pricingResolver, etc. i was trying to make it operate like your contracts with the 721store. so i am clear, you guys want me to do it as you have commented above and NOT how i was doing it, or does it matter? how opinionated if its cleaned up? 

because the NFT factory for just tokens outside the system just paying into the treasury is going to be its own thing, in its own repo. i did want to have the code all feel and look the same even if they were in different repos.

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+    @notice
+    NFT mint cap as part of this configuration.
+  */
+  uint256 private _maxSupply;
+
+  /**
+    @notice
+    Amount of NFTs minted thus far.
+  */
+  uint256 private _distributedSupply;
```
drgorillamd, line 88: double use with _nextTokenId no?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+      _data.amount.value >= _minContribution.value &&
+      _data.amount.currency == _minContribution.currency
+    ) {
+      
+      _mint(_data.beneficiary, _nextTokenId);
+
+      ++_supply;
+      ++_nextTokenId;
+
+      ++_distributedSupply;
```
drgorillamd, line 202: iiuc, you can avoid having both _nextTokenId and distributedSupply (would save one sstore on every mint) -> isn't _nextTokenId == distributedSupply? Or is there a scenario where it wouldn't be the case that I'm missing?

drgorillamd: same with supply actually, looks like Rari burn doesn't change it

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  /**
+    @notice
+    Owner-only function to burn an address' token.
+
+    @param _account The address which owns the token to be burned.
+    @param _tokenId The tokenId which will be burned..
+   */
+  function burn(address _account, uint256 _tokenId) external override onlyOwner {
```
drgorillamd, line 362: Do we really need this':)?

contracts/JBNFTRewardDataSourceDelegate.sol
drgorillamd: and with _supply (since no change when burning)

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+  // ------------------- IJBFundingCycleDataSource --------------------- //
+  //*********************************************************************//
+
+  function payParams(JBPayParamsData calldata _data) 
+    external 
+    view 
+    override 
+    returns (uint256 weight, string memory memo, IJBPayDelegate delegate) 
+  {
+    return (0, _data.memo, IJBPayDelegate(address(this)));
```
drgorillamd, line 168: perhaps a more universal approach could be fun (ie a custom weight passed in constructor, and defaulting to _data.weight if type.max for instance)?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  function redeemParams(JBRedeemParamsData calldata _data)
+    external
+    pure
+    override
+    returns (uint256 reclaimAmount, string memory memo, IJBRedemptionDelegate delegate)
+  {
+    return (0, _data.memo, IJBRedemptionDelegate(address(0)));
```
drgorillamd, line 177: same for weight?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+  // ------------------------ IJBPayDelegate --------------------------- //
+  //*********************************************************************//
+
+  function didPay(JBDidPayData calldata _data) 
+    external 
+    override 
+  {
+    if (!_directory.isTerminalOf(_projectId, IJBPaymentTerminal(msg.sender))) revert INVALID_PAYMENT_EVENT();
+
+    if (_distributedSupply == _maxSupply) return;
```
drgorillamd, line 190: with the weight of 0 supra and this check, contributors would then pay without anything in return (no more NFT and no projct token to at least claim back their eth) -> per design?

contracts/JBNFTRewardDataSourceDelegate.sol
```diff
@@ -0,0 +1,401 @@
...
+  function didPay(JBDidPayData calldata _data) 
+    external 
+    override 
+  {
+    if (!_directory.isTerminalOf(_projectId, IJBPaymentTerminal(msg.sender))) revert INVALID_PAYMENT_EVENT();
+
+    if (_distributedSupply == _maxSupply) return;    
+
+    if (
+      _data.amount.value >= _minContribution.value &&
```
drgorillamd, line 193: or mint multiple of _minContribution? the '>' can be dangerous for people not realizing what would happen if they pay 10eth (for a min of 1)  -  rekt:)

contracts/interfaces/IJBController/1.sol
```diff
@@ -172,7 +172,7 @@ interface IJBController is IERC165 {
     uint256 _projectId,
     string calldata _name,
     string calldata _symbol
-  ) external returns (IJBToken token);
+  ) external returns (IJBToken);
```
drgorillamd, line 175: why?

contracts/system_tests/TestNFTRewardsFlow.sol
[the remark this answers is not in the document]
drgorillamd: readibility increases if you assign a new value only to what needs to be changed, not redeclaring the whole struct

contracts/system_tests/helpers/TestBaseWorkflow.sol
```diff
@@ -16,6 +16,7 @@ import '../../JBProjects.sol';
 import '../../JBSplitsStore.sol';
 import '../../JBToken.sol';
 import '../../JBTokenStore.sol';
+import '../../JBNFTRewardDataSourceDelegate.sol';
```
drgorillamd, line 19: don't you mind moving this import into the test itself? Not part of the base per se

contracts/system_tests/TestNFTRewardsFlow.sol
```diff
@@ -0,0 +1,171 @@
...
+      useTotalOverflowForRedemptions: false,
+      useDataSourceForPay: true,
+      useDataSourceForRedeem: false,
+      dataSource: address(_jbNFTRewardDataSourceDelegate)
+    });
+
+    evm.warp(block.timestamp + 10000);
+
+    evm.prank(_projectOwner);
+    _controller.reconfigureFundingCyclesOf(
```
drgorillamd, line 128: why a reconfiguration here?

test/jb_nft_reward_datasource_delegate/owner_mint_burn.js
```diff
@@ -0,0 +1,107 @@
+import { expect } from 'chai';
```
drgorillamd, line 1: this test is skipped -> add .test in the filename

test/jb_nft_reward_datasource_delegate/did_pay.test.js
```diff
@@ -0,0 +1,191 @@
...
+      payer: beneficiary.address,
+      projectId: PROJECT_ID,
+      currentFundingCycleConfiguration: 0,
+      amount: { token: ethToken, value: ETH_TO_PAY, decimals: 18, currency: CURRENCY_ETH },
+      projectTokenCount: 0,
+      beneficiary: beneficiary.address,
+      preferClaimedTokens: true,
+      memo: '',
+      metadata: '0x42'
+    })).to.emit(jbNFTRewardDataSource, 'Transfer').withArgs(ethers.constants.AddressZero, beneficiary.address, 0);
```
drgorillamd, line 67: perhaps test balance pre/post-mint?

test/jb_nft_reward_datasource_delegate/did_pay.test.js
```diff
@@ -0,0 +1,191 @@
...
+      payer: beneficiary.address,
+      projectId: PROJECT_ID,
+      currentFundingCycleConfiguration: 0,
+      amount: { token: ethToken, value: ETH_TO_PAY, decimals: 18, currency: CURRENCY_ETH },
+      projectTokenCount: 0,
+      beneficiary: beneficiary.address,
+      preferClaimedTokens: true,
+      memo: '',
+      metadata: '0x42'
+    })).not.to.emit(jbNFTRewardDataSource, 'Transfer');
```
drgorillamd, line 95: That's what I meant in contract's comment -> no mint, no project token, just some eth gone :D

test/jb_nft_reward_datasource_delegate/did_pay.test.js
```diff
@@ -0,0 +1,191 @@
...
+      projectTokenCount: 0,
+      beneficiary: beneficiary.address,
+      preferClaimedTokens: true,
+      memo: '',
+      metadata: '0x42'
+    })).to.be.revertedWith('INVALID_PAYMENT_EVENT()');
+
+  });
+
+  it(`Tests for unsupported pay functions`, async function () {
```
drgorillamd, line 131: Those are returning values (and are therefore 'supported' since they're pointing to the correct delegate to use later on in pay(..)) + returning a weight of 0 (ie 0 project token minted)

contracts/JBNFTRewardDataSourceDelegate.sol
drgorillamd: + those are capped to max supply -> can be unchecked for some more gas golfing

PR #262, opened 2022-05-16
added param

mejango:
Added a param in internal fn for inheriters to specify where to send tokens if not specified in splits.

Useful for JBMarket here https://github.com/jbx-protocol/juice-market/blob/9fa048c3f0ddacc81de7a60ee04dc3cf27a14a30/contracts/JBMarket.sol#L406

needs tests @drgorillamd

### DIFF
--- a/contracts/JBETHERC20SplitsPayer.sol
+++ b/contracts/JBETHERC20SplitsPayer.sol
@@ -151,7 +151,7 @@ contract JBETHERC20SplitsPayer is IJBSplitsPayer, JBETHERC20ProjectPayer, Reentr
       JBTokens.ETH,
       address(this).balance,
       18, // decimals.
-      msg.sender
+      defaultBeneficiary != address(0) ? defaultBeneficiary : msg.sender
     );
 
     // If there is no leftover amount, nothing left to pay.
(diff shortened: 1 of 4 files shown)


contracts/JBETHERC20SplitsPayer.sol
```diff
@@ -150,7 +150,8 @@ contract JBETHERC20SplitsPayer is IJBSplitsPayer, JBETHERC20ProjectPayer, Reentr
       defaultSplitsGroup,
       JBTokens.ETH,
       address(this).balance,
-      18 // decimals.
+      18, // decimals.
+      msg.sender
```
drgorillamd, line 154: Should we have a default value here too? (ie `defaultBeneficiary == address(0) ? msg.sender : defaultBeneficiary` )

drgorillamd: (and same in pay/addToBalanceOf)

mejango: nice. using the already-available `defaultBeneficiary`

drgorillamd (approved):

PR #264, opened 2022-05-23
v5.0.0: Move dependencies to devDependencies

aeolianeth:
Consumers of the npm package are only concerned with the `deployments/` directory, and don't need any of the currently listed packages in `dependencies`.

All of these can be moved to `devDependencies`. The result is that `npm install @jbx-protocol/contracts-v2` won't install any dependencies, saving a bunch of hard disk space and install time.


package.json
```diff
@@ -7,32 +7,30 @@
     "type": "git",
     "url": "https://github.com/jbx-protocol/juice-contracts-v2"
   },
-  "version": "4.0.0",
+  "version": "5.0.0",
```
aeolianeth, line 10: I think a major version is the safest.

 `juice-interface` still compiles from this PR because we define all the necessary dependencies ourselves, but other consumers might have been relying on `ethers` from `@jbx-protocol/contracts-v2`; if they installed `@jbx-protocol/contracts-v2` again and didn't have their version locked, it would break their project.

mejango:
@aeolianeth mind removing the version bump from this one? I'll bump the version after redeploy.

package.json
drgorillamd, line 10:
```suggestion
  "version": "4.0.0",
```

drgorillamd (changes requested):

drgorillamd (approved):

mejango (approved):

PR #266, opened 2022-05-24
done

mejango:
two things:

- JBSplitStore.set now takes an array for grouped splits.
- private fns and properties in terminals and controllers are now internal.

need tests.


drgorillamd (approved):
lfg

PR #279, opened 2022-06-10
payDelegate can now receive funds directly from the terminal

mejango:
in progress. 

Proposed diff for a v2.1 terminal that projects can migrate to.

No bug fixes, only features:

- PayDelegate is payable, data sources can specify how much of the paid amount should be sent to the pay delegate from the terminal. 

These changes shouldn't land on main. they should be separated into its own repo. Looking at the diff here is useful to see if its worth doing.


mejango:
This aint pressing but it's real interesting when considering fee extensions, apple juice, or any funds routing extensions.

mejango:
folded into #285

PR #284, opened 2022-07-09
Reordered (interface) inheritance

xBA5ED:
Reordered inheritance to always have interfaces at the end.
[More info regarding why](https://discord.com/channels/775859454780244028/995010619273138246/995075636685193218)


mejango (approved):
hell yeah. lgtm

PR #287, opened 2022-07-17
done

mejango:
I think this JBControllerV2.1 is best served in a stand-alone repo with tests ported over instead of landed on `main`. but here it is for review. 

Needs test adjustments and documentation adjustments.

### DIFF
--- a/contracts/JBController.sol
+++ b/contracts/JBController.sol
@@ -47,6 +47,7 @@ contract JBController is JBOperatable, ERC165, IJBController, IJBMigratable {
   error INVALID_OVERFLOW_ALLOWANCE_CURRENCY();
   error INVALID_REDEMPTION_RATE();
   error INVALID_RESERVED_RATE();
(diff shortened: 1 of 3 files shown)


mejango:
@drgorillamd mind helping moving the JBController and tests into a new juice-jbcontroller-v2.1 repo? unit tests are in hardhat so there may be a little finicking. maybe first commit to `main` should be the unchanged v2 controller, and first PR are the changes.

drgorillamd:
> @drgorillamd mind helping moving the JBController and tests into a new juice-jbcontroller-v2.1 repo? unit tests are in hardhat so there may be a little finicking. maybe first commit to `main` should be the unchanged v2 controller, and first PR are the changes.

Hmm, let name the repo as you suggest but perhaps just fork the v2, so we can easily reuse the forge tests too?
If not, we can keep them but need to fix the imports

drgorillamd:
See v3 repo

PR #285, opened 2022-07-13
first pass of all ideal changes

mejango:

PR #289, opened 2022-07-23
adjusted split store given feedback

mejango:
Adjusted splits store given code4rena feedback.

JBSplitStoreV2_1 should be used in JBControllerV2_1 and JBPayoutRedemptionPaymentTerminalV2_1


mejango:
folded into #285

PR #297, opened 2022-09-13
fix: solc ^0.8.0 in non-logic contracts

drgorillamd:
Improve compatibility with other repos reusing our interfaces or constant, by setting solc to ^0.8.0 (no version change for us as main contracts stay ==0.8.16)


mejango (approved):

drgorillamd:
Improve compatibility with other repos reusing our interfaces or constant, by setting solc to ^0.8.0 (no version change for us as main contracts stay ==0.8.16)

